PR #3, opened 2017-07-07
Support versions of Go down to Go1.6

### LINKED ISSUE #2: Compile under Go 1.6
For now, we're only worrying about current Go version, but since we'd like to use this for App Engine-targeted packages as well, it would be good to support Go 1.6.

dsnet:
Avoid any features introduced in newer versions such as:
* testing.T.Run
* sort.Slice
* sort.SliceStable
* sort.IsSliceSorted

Fixes #2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: false
 language: go
-# TODO(light): Enable Go 1.6 (#2).
 go:
+- 1.6
 - 1.x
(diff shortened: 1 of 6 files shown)


dsnet:
Nothing in the *.go changed since you last saw. Can you just check the travis file?

zombiezen:
It's probably sufficient to just have:

```yaml
go:
- 1.6
- 1.x
```

PR #7, opened 2017-07-14
-Better and elegantly connect strings.

wgliang:
Better and elegantly connect strings using fmt.Sprintf


wgliang:
I signed it! @googlebot

PR #9, opened 2017-07-14
Add diffing abilities to reporting

dsnet:
This adds an internal package with the following API:
    func Difference(nx, ny int, f EqualFunc) (eq bool, es EditScript)
    type EditScript []EditType
    type EditType int8
        const Identity EditType = iota ...
    type EqualFunc func(ix int, iy int) Result
    type Result struct{ ... }

The cmp.SliceIndex type has a new method SplitKeys to indicate that
elements are being compared in x and y are in different indexes.

The debug visualizer can be run:
	go test -tags=debug -v github.com/google/go-cmp/cmp/internal/diff

Reviewed-By: dneil@google.com

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -29,6 +29,8 @@ package cmp
 import (
 	"fmt"
 	"reflect"
+
+	"github.com/google/go-cmp/cmp/internal/diff"
 )
 
 // BUG: Maps with keys containing NaN values cannot be properly compared due to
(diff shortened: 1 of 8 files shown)


cmp/example_test.go
```diff
@@ -147,6 +147,7 @@ func ExampleOption_equalEmpty() {
 	// false
 }
 
+
```
shurcooL: Is this change intentional?

It looks like it makes the file no longer `gofmt`ed.

dsnet: Good catch. Fixed.

shurcooL: Would you like a PR that adds a `gofmt` check to Travis CI, so this type of issue can be caught automatically by CI in all future contributions?

dsnet: That would be great, thank you.

PR #10, opened 2017-07-14
Add helpful suggestions to panic message

dsnet:
Users find it unhelpful to simply panic saying that unexported fields
cannot be handled. Augment the panic message by pointing the user
to functions that can be used to address the issue.

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -67,7 +67,7 @@ import (
 //
 // Structs are equal if all of their fields are equal. If a struct contains
 // unexported fields, Equal panics unless the AllowUnexported option is used or
-// an Ignore option ignores that field.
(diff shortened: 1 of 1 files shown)


cmp/compare.go
```diff
@@ -279,7 +279,8 @@ func (s *state) tryOptions(vx, vy *reflect.Value, t reflect.Type) bool {
 	// are either exported or can be forcibly exported.
 	if sf, ok := s.curPath[len(s.curPath)-1].(*structField); ok && sf.unexported {
 		if !sf.force {
-			panic(fmt.Sprintf("cannot handle unexported field: %#v", s.curPath))
+			help := "consider using AllowUnexported or cmpopts.IgnoreUnexported"
```
bcmills: `const help =`?

PR #15, opened 2017-07-18
Move general reflect logic to internal/value

dsnet:
This internal package has the API:
	Format(reflect.Value, useStringer bool) string
	SortKeys(vs []reflect.Value) []reflect.Value

While moving Format and SortKeys, we also tweak the formatting logic
to print the type on named primitive types.


cmp/internal/value/format.go
```diff
@@ -0,0 +1,258 @@
+// Copyright 2017, The Go Authors. All rights reserved.
```
bcmills: Is there some way to convince Git and/or GitHub to show this file as a diff on the file it was extracted from?

It looks like this file should be a pretty small delta, but that's hard to see visually in this code review tool...

shurcooL: For GitHub, not that I know of.

For git, I think you can adjust the `
--find-renames[=<n>]` (aka `-M[<n>]`) flag of `git diff`, see https://git-scm.com/docs/git-diff. You may be able to get it to do what you want.

Alternatively, you can force seeing a diff between specific files by copying/pasting the two files into a diff tool, e.g., https://www.quickdiff.com/. (It'll look like [this](https://user-images.githubusercontent.com/1924134/28347042-c09f9170-6c02-11e7-99e9-ac1e71d6c20a.png).)

dsnet: Unfortunately, it seems that git does not have a concept of "copy this file with history".

https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history

PR #16, opened 2017-07-18
Travis: Run go vet; use race detector during tests.

### LINKED ISSUE #13: Perform additional checks in Travis CI.
@dsnet, in https://github.com/google/go-cmp/pull/9#discussion_r127847401, I asked if you wanted a PR that adds a `gofmt` check to Travis CI, and you said "that would be great". So I sent #12 for that.

However, I'd like to propose performing additional checks in CI. After years of iterating on it, I've arrived at the following `.travis.yml` script, which I [generate](https://github.com/shurcooL/cmd/blob/3775d1a67e78fba6a8380e30337340465a0b30e3/gorepogen/template.go#L50-L66) for all my Go repositories:

```yml
sudo: false
language: go
go:
  - 1.x
  - master
matrix:
  allow_failures:
    - go: master
  fast_finish: true
install:
  - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
script:
  - go get -t -v ./...
  - diff -u <(echo -n) <(gofmt -d -s .)
  - go tool vet .
  - go test -v -race ./...
```

Here's an explanation of its features over the minimal `.travis.yml` change I sent in #12 that adds the `gofmt` only:

- Use `-s` flag with `gofmt` to ensure code simplifications are made.
- Use `-race` flag with `go test` to enable race detector.
- Run `go vet` on the directory to catch `vet` issues.
- Run a fast-finish allowed failure build on Go tip version. Allowed failure means a temporary issue in Go tip won't fail the build, but one can still manually see if there are some real failures in that version by manually checking Travis builds occasionally. Fast-finish just means the overall build status is computed faster, without waiting for allowed failure steps (I'm not really sure why it's an option instead of just being a default). See more info about it at https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/.
- Move `go get -t -v ./...` step from `install` step to `script`. This one is very minor and more of a personal preference, but it makes it so that if there's a build failure in any of the dependencies, the build appears as a red "failure", rather than a gray "error".

@dsnet Let me know if you'd like any or all of the above enhancements, I'm happy to send a followup PR.

shurcooL:
This change augments the Travis CI build to perform:

-	Include `-s` (simplify) flag when checking that all files follow `gofmt` style.
-	Check that `go vet` does not report any problems.
-	Use race detector when running tests, to help catch data races.
-	Include a fast-finish allowed failure build on `master` (tip). Due to a chance of false positives, failures on tip are not critical, but manually looking over them can help spot issues either in this project, or in the tip version of Go.

Resolves #13. /cc @dsnet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,14 @@ language: go
 go:
   - 1.x
   - 1.6
+  - master
+matrix:
+  allow_failures:
+    - go: master
+  fast_finish: true
+install:
+  - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (it is intended for this package to have no dependencies other than the standard library).
 script:
-  - diff -u <(echo -n) <(gofmt -d .)
-  - go test -v ./...
+  - diff -u <(echo -n) <(gofmt -d -s .)
+  - go tool vet .
+  - go test -v -race ./...
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,11 @@ sudo: false
 language: go
 go:
   - 1.x
-  - 1.6
   - master
 matrix:
+  include:
+    - go: 1.6.x
+      script: go test -v -race ./...
   allow_failures:
     - go: master
   fast_finish: true
--- a/cmp/cmpopts/sort_go18.go
+++ b/cmp/cmpopts/sort_go18.go
@@ -15,8 +15,8 @@ const hasReflectStructOf = true
 
 func mapEntryType(t reflect.Type) reflect.Type {
 	return reflect.StructOf([]reflect.StructField{
-		reflect.StructField{Name: "K", Type: t.Key()},
-		reflect.StructField{Name: "V", Type: t.Elem()},
+		{Name: "K", Type: t.Key()},
+		{Name: "V", Type: t.Elem()},
 	})
 }
 
--- a/cmp/cmpopts/util_test.go
+++ b/cmp/cmpopts/util_test.go
@@ -200,14 +200,14 @@ func TestOptions(t *testing.T) {
 	}, {
 		label: "SortMaps",
 		x: map[MyTime]string{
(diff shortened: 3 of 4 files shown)


.travis.yml
```diff
@@ -3,6 +3,14 @@ language: go
...
+matrix:
+  allow_failures:
+    - go: master
+  fast_finish: true
+install:
+  - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (it is intended for this package to have no dependencies other than the standard library).
 script:
-  - diff -u <(echo -n) <(gofmt -d .)
-  - go test -v ./...
+  - diff -u <(echo -n) <(gofmt -d -s .)
```
dsnet: I suspect that Travis will complain shortly because of the '-s' flag :)

Feel free to commit the simplified Go files in the same PR.

shurcooL:
Everything is succeeding on Go 1.8 and tip, but `go tool vet .` is failing on Go 1.6.

In general, I think the best thing to do is to run `go vet` and similar checks on latest version of Go only, and only perform `go test` on older versions.

I'll try to make that change.

dsnet:
SGTM. I can't wait until I can nuke Go1.6 support...

dsnet:
Does this hack work?
```
if [[ $(go version) != *"go1.6"* ]]; then go tool vet .; fi
```

shurcooL:
Yes, I believe that (or something along those lines) would work.

You could make it a little nicer by using the `TRAVIS_GO_VERSION` env var which includes version only without prefix/suffix, as described here https://docs.travis-ci.com/user/languages/go/#Environment-Variable. [Many people](https://github.com/search?q=TRAVIS_GO_VERSION&type=Code) seem to use it in a similar way as you suggested.

However, I believe I know a slightly cleaner way, I'll see if I can make it work here.

shurcooL:
PTAL.

dsnet:
This is great! Learned a few things about travis too! Thank you.

PR #19, opened 2017-07-19
Clarify documentation regarding empty slices and maps

### LINKED ISSUE #18: nil slice/map should compare equal to empty slice/map
The documentation states:

	Slices and arrays are equal if they have the same length and the elements
	at each index are equal. Maps are equal if their keys are exactly equal
	(according to the == operator) and the corresponding elements for each
	key are equal.

I read this as implying that this program would print `true true`, but
instead it prints `false false`.

	package main
	
	import (
		"log"
	)
	
	func main() {
		log.Println(
			cmp.Equal(map[int]string{}, map[int]string(nil)),
			cmp.Equal([]int{}, []int(nil)),
		)
	}

dsnet:
State that a non-nil empty slice is not equal to a nil slice.
Suggest use of cmpopts.EquateEmpty to handle this case.

Fixes #18

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -69,12 +69,13 @@ import (
 // Structs are equal if all of their fields are equal. If a struct contains
 // unexported fields, Equal panics unless the AllowUnexported option is used or
 // an Ignore option (e.g., cmpopts.IgnoreUnexported) ignores that field.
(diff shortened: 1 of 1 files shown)


dsnet:
\cc @rogpeppe

PR #20, opened 2017-07-20
Expand dynamic checks of options

dsnet:
Add checks for mutations of the input value and non-deterministic
transformer outputs.

A prior approach (not shown in this commit) attempted to detect mutations of
the input by hashing the inputs before and after calling the option functions.
However, we deemed this approach too restrictive since there are legitimate
cases where some types lazily update their values upon first access
(for example, accessing extensions on proto2 messages). This would cause the
hashes to mismatch and provide a false-positive panic.

In this approach, we accept the reality that lazy evaluators exist, but make
the assumption that any lazy evaluators are protected by a mutex of some sort.
Thus, we simply run the option function twice in parallel, and let the race
detector detect mutations on the input. We intentionally run the options
in a function named detectRaces to make it obvious from the stack trace
what's going wrong.

For the detection of deterministic output of Transformers, we call a special
version of s.compareAny so that the outputs may be compared according to the
semantics of what the user provided. The specialized wrapper s.statelessCompare
preserves the comparison state such as the current result and the reporter.

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -106,25 +106,20 @@ func Diff(x, y interface{}, opts ...Option) string {
 }
 
 type state struct {
-	result  diff.Result // The current result of comparison
-	curPath Path        // The current path in the value tree
+	// These fields represent the "comparison state".
+	// Calling statelessCompare must not result in observable changes to these.
+	result   diff.Result // The current result of comparison
+	curPath  Path        // The current path in the value tree
+	reporter reporter    // Optional reporter used for difference formatting
 
-	// dsCheck tracks the state needed to periodically perform checks that
-	// user provided func(T, T) bool functions are symmetric and deterministic.
-	//
-	// Checks occur every Nth function call, where N is a triangular number:
-	//	0 1 3 6 10 15 21 28 36 45 55 66 78 91 105 120 136 153 171 190 ...
-	// See https://en.wikipedia.org/wiki/Triangular_number
-	//
-	// This sequence ensures that the cost of checks drops significantly as
-	// the number of functions calls grows larger.
-	dsCheck struct{ curr, next int }
+	// dynChecker triggers pseudo-random checks for option correctness.
+	// It is safe for statelessCompare to mutate this value.
+	dynChecker dynChecker
 
 	// These fields, once set by processOption, will not change.
 	exporters map[reflect.Type]bool // Set of structs with unexported field visibility
 	optsIgn   []option              // List of all ignore options without value filters
 	opts      []option              // List of all other options
-	reporter  reporter              // Optional reporter used for difference formatting
 }
 
 func newState(opts []Option) *state {
@@ -172,6 +167,24 @@ func (s *state) processOption(opt Option) {
 	default:
 		panic(fmt.Sprintf("unknown option %T", opt))
 	}
+}
+
+// statelessCompare compares two values and returns the result.
+// This function is stateless in that it does not alter the current result,
+// or output to any registered reporters.
+func (s *state) statelessCompare(vx, vy reflect.Value) diff.Result {
+	// We do not save and restore the curPath because all of the compareX
+	// methods should properly push and pop from the path.
+	// It is an implementation bug if the contents of curPath differs from
+	// when calling this function to when returning from it.
+
+	oldResult, oldReporter := s.result, s.reporter
+	s.result = diff.Result{} // Reset result
+	s.reporter = nil         // Remove reporter to avoid spurious printouts
+	s.compareAny(vx, vy)
+	res := s.result
+	s.result, s.reporter = oldResult, oldReporter
+	return res
 }
 
 func (s *state) compareAny(vx, vy reflect.Value) {
@@ -330,7 +343,7 @@ func (s *state) applyFilters(vx, vy reflect.Value, t reflect.Type, opt option) b
 		}
 	}
 	for _, f := range opt.valueFilters {
-		if !t.AssignableTo(f.in) || !s.callFunc(f.fnc, vx, vy) {
+		if !t.AssignableTo(f.in) || !s.callTTBFunc(f.fnc, vx, vy) {
 			return false
 		}
 	}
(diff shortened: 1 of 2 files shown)


cmp/compare.go
```diff
@@ -361,26 +352,55 @@ func (s *state) tryMethod(vx, vy reflect.Value, t reflect.Type) bool {
...
+		return f.Call([]reflect.Value{v})[0]
+	}
+
+	// Run the function twice and ensure that we get the same results back.
+	// We run in goroutines so that the race detector (if enabled) can detect
+	// unsafe mutations to the input.
+	c := make(chan reflect.Value)
+	go detectRaces(c, f, v)
+	want := f.Call([]reflect.Value{v})[0]
+	if got := <-c; !value.Equal(got, want) {
```
bcmills: This doesn't seem right to me: shouldn't we compare `got` and `want` using the notion of equality defined by the user-provided options?

It seems like `got` and `want` can differ arbitrarily in terms of internal representation — for example, one could represent a lazily-evaluated term before evaluation, while the other could represent a (memoized or cached) evaluated version. We only really care about the algebraic properties of the value with respect to the rest of the `cmp.Equal` or `cmp.Diff` call, not other internals.

cmp/compare.go
```diff
@@ -361,26 +352,55 @@ func (s *state) tryMethod(vx, vy reflect.Value, t reflect.Type) bool {
...
 	}
-	s.dsCheck.curr++
-	return got
+	return want
+}
+
+func detectRaces(c chan<- reflect.Value, f reflect.Value, vs ...reflect.Value) {
+	var ret reflect.Value
+	defer func() {
+		recover() // Ignore panics, let the other call to f panic instead
```
bcmills: > `let the other call to f panic instead`

What happens if the other call doesn't panic?

(For example, what if they're making [fragile assumptions about goroutines](http://godoc.org/github.com/jtolds/gls) and only the off-goroutine ever panics?)

cmp/compare.go
dsnet: In an implementation of lazy-evaluated value, wouldn't the overall structure of the value returned still be the same (i.e., the returned value in both cases are in the lazy form)? In the case of a memoized value, would it be the exact same value returned from each call to the transformer? Either way, our loose definition of DeepEqual would work. For the detection of mutation on the *input*, extensions on proto messages was a realistic case where it would fail. What's a realistic case where this check on the *output* would fail? 

If possible I want to avoid plumbing this through all the logic of `compareAny`, which would require needing to ensure the `curPath` is updated, that we don't accidentally output to the reporter, and also that we don't clobber any other state.

cmp/compare.go
dsnet: Then the return value of this is bogus, and the logic above will still panic with the non-deterministic message (which is still true).

If you feel it's worth it, I can add logic to capture the panic object and re-panic it in the original goroutine. The reason that I don't let the panic occur in this routine is that it makes testing for panics (in the cmp and cmpopts unit tests) hard since you can't recover a panic from a different goroutine.

cmp/compare.go
bcmills: > What's a realistic case where this check on the _output_ would fail?

I could envision a memo cell implemented as something like:

```go
type memo struct {
    ready <-chan struct{}
    v interface{}
}

func (m *memo) Set(f func() interface{}) {
    if m.ready != nil {
        return
    }
    m.ready = make(chan struct{})
    go func() {
        m.v = f()
        close(m.ready)
    }()
}

func (m *memo) Get() interface{} {
    <-m.ready
    return m.v
}
```

...where application code considers two memo cells to be equal if their `Get` methods return equal values.

For such a memo cell, you could envision the Transformer accepting a pointer to a larger struct and returning a pointer to a field of type `memo` within it. Calling `value.Equal` on such a memo cell would produce a data race instead of a correct comparison.

(I can imagine a similar scenario occurring with a `*sync.Map`: for example, `(encoding/json).typeEncoder` uses a similar trick to handle recursive types, but it happens to block until the final encoder for the type has been set so it likely would not fail in this case.)

cmp/compare.go
bcmills: > The reason that I don't let the panic occur in this routine is that it makes testing for panics (in the cmp and cmpopts unit tests) hard since you can't recover a panic from a different goroutine.

You can always execute the panicking code in a subprocess.  :)

(But do we really need to test what happens when user code panics anyway?)

cmp/compare.go
dsnet: So, assuming the transformer does something like:
```go
m := memo{ready: make(chan struct{})}
m.Set(func() interface{} {
    return computeExpensiveValue()
})
return m
```

The equality will still report true since the `ready` chan will both be non-nil and the `v` value will both be nil. It won't be able to detect bad implementations of functions passed to `memo.Set`, but at least it doesn't trigger a false positive, which I'm okay with.

bcmills: > the `ready` chan will both be non-nil and the `v` value will both be nil.

In this case, the `v` values may or may not be nil, but reading them without first receiving from the channel is a data race.

I agree that false negatives are fairly benign, but false data races are not, and I'm not comfortable restricting transformers to return only types that do not have internal synchronization.

dsnet: I tried prototyping something that would use `compareAny` directly. One complication is that the default implementation of `compareAny` is not symmetric, causing false-positives when `NaNs` are involved. We can special-case NaNs to be equal to each other, but it's adding complexity to the logic.

dsnet: Here's an idea. We can trivially detect non-reflexive comparisons by simply checking if 
`compareAny(got, got) == true` and only doing `compareAny(got, want)` if true.

bcmills: It's probably less expensive to do the `(got, want)` check first, since it will ~always succeed. Then the condition is `compareAny(got, want) || !compareAny(want, want)`.

cmp/compare_test.go
```diff
@@ -334,6 +335,41 @@ root:
...
+		label: label,
+		x:     make([]int, 10),
+		y:     make([]int, 10),
+		opts: []cmp.Option{
+			cmp.Transformer("", func(x int) float64 {
+				return math.NaN()
+			}),
+		},
+		wantDiff: `
+{[]int}:
```
dsnet: Hmm interesting. Irrelevant to this pull request, but this prints with `0`s rather than `NaN`s because when printing the full slice it doesn't know that transformations has happened lower in the tree. I'm not sure what the right behavior in this situation is.

Issue #17, opened 2017-07-18
Should Transformers be recursive?


dsnet:
The current implementation of cmp.Equal recursively applies any Transformers that remain after applying all the filters. For a trivial Transformer of the type `func(T) T`, this will be recursively apply to the output since it will always match. For this reason, the [`cmpopts.SortSlices`](https://github.com/google/go-cmp/blob/e6ad8e8c6f4959eb2ffe240ab826c76b2fb855fd/cmp/cmpopts/sort.go#L48-L52) is wrapped with a `FilterValues` that acts as a "base-case" to prevent infinite recursion. One alteration to the algorithm is to say that a given Transformer cannot be recursively applied on a value that was transformed at some point by that very Transformer.

The major advantage of non-recursive Transformers is:
* A naive use of `func(T) T` works as expected without any need for a "base-case" filter.

The disadvantages of non-recursive Transformer is:
* The additional documentation needed to explain the special-case where Transformers are *not* executed. The implementation of this is essentially a set of "spent" Transformers, where transformers are added to the set when descending down the tree, and popped when ascending up the tree.
* In some rare use-cases you may actually want Transformers to be recursive. Imagine you have a non-binary tree, where each node is essentially a set of children. When comparing this data-structure, you may want to sort each node using a transformer. You would need the property of recursive transformers to ensure each node may get transformed.
* The need for "base-case" filters to prevent recursive cycles.

Thoughts? I do see the value in non-recursive Transformers. It simplifies use of them, but prevents some very rare use-cases.

\cc @bcmills @jba

rogpeppe:
FWIW my first experiment with this package, was to test that transformers were repeatedly applied, as I think that makes things more scalable, always transforming towards the goal of comparable types.

I think the current implementation is both intuitive and useful.

jba:
I'm probably confused, but as an unlikely working hypothesis I'm going to assume instead that both of you are confused.

Here's how I see it: At a given call of cmp.Equal, you have a value `x` (well, actually, two values `x` and `y`, but they behave independently w.r.t. transformers) and a set `S` of transformers that have passed the filters.  You then do the following repeatedly:
1. Choose a transformer `t` in `S` that can be applied to `x`. If none, terminate. If more than one, fail.
2. Apply `t` to `x`, giving a new `x`.
3. Remove `t` from `S`.

Step 3 ensures that this algorithm will terminate.

[...]

PR #26, opened 2017-07-25
Use policy where Transformers are not recursively called

dsnet:
DO NOT SUBMIT

Updates #17


dsnet:
This approach is inferior to #29.

PR #29, opened 2017-07-27
Add implicit filter to Transformers

### LINKED ISSUE #17: Should Transformers be recursive?

### LINKED ISSUE #51: Make it easier to do same-type transforms
I wanted to treat `time.Time`s as rounded to the nearest microsecond. I first tried a transform, but it got into an infinite loop (as advertised). I had to add a FilterPath in front of it. This turned a one-line option into a ~10-line option.

So how about adding something like this to cmpopts:
```
// TransformOnce applies f, which should be of type func(T) T, to values of type T.                                                                                                            
// Unlike Transform, f is applied only once to each original value.                                                                                                                            
func TransformOnce(name string, f interface{}) cmp.Option {
    return cmp.FilterPath(func(p cmp.Path) bool {
        if len(p) == 0 {
            return true
        }
        xform, ok := p[len(p)-1].(cmp.Transform)
        if !ok {
            return true
        }
        return xform.Name() != name
    }, cmp.Transformer(name, f))
}
```

I know we discussed this earlier, but I still don't really understand why this isn't the default behavior of transformers. 

/cc @bcmills

dsnet:
Declaring a transformer "func(T) T" where T is a concrete type
is a common transformation. However, this is currently problematic
as the transformation now infinitely applies to itself recursively.

In order to allow this form of transformation, add a simple
(but subtle) filter to Transformers where a Transformer can only
apply if that specific Transformer does not already exist within
the tail of the current Path since the last non-Transform step.

This rule does not prevent more advance usages of Transformers
where the user *does* want the Transformer to apply recursively
to previously transformed output, since those situations are
almost always followed by some path step that is *not* a
transformation (e.g., pointer indirect, slice indexing, etc).

Fixes #17
Fixes #51

### DIFF
--- a/cmp/options.go
+++ b/cmp/options.go
@@ -215,9 +215,12 @@ func (invalid) apply(s *state, _, _ reflect.Value) bool {
 // The transformer f must be a function "func(T) R" that converts values of
 // type T to those of type R and is implicitly filtered to input values
 // assignable to T. The transformer must not mutate T in any way.
-// If T and R are the same type, an additional filter must be applied to
-// act as the base case to prevent an infinite recursion applying the same
-// transform to itself (see the SortedSlice example).
+//
+// To help prevent some cases of infinite recursive cycles applying the
+// same transform to the output of itself (e.g., in the case where the
+// input and output types are the same), an implicit filter is added such that
+// a transformer is applicable only if that exact transformer is not already
+// in the tail of the Path since the last non-Transform step.
 //
 // The name is a user provided label that is used as the Transform.Name in the
 // transformation PathStep. If empty, an arbitrary name is used.
@@ -248,7 +251,14 @@ type transformer struct {
 
 func (tr *transformer) isFiltered() bool { return tr.typ != nil }
 
-func (tr *transformer) filter(_ *state, _, _ reflect.Value, t reflect.Type) applicableOption {
+func (tr *transformer) filter(s *state, _, _ reflect.Value, t reflect.Type) applicableOption {
+	for i := len(s.curPath) - 1; i >= 0; i-- {
+		if t, ok := s.curPath[i].(*transform); !ok {
+			break // Hit most recent non-Transform step
+		} else if tr == t.trans {
+			return nil // Cannot directly use same Transform
+		}
+	}
 	if tr.typ == nil || t.AssignableTo(tr.typ) {
 		return tr
 	}
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -1633,15 +1633,11 @@ func (gs germSorter) Swap(i, j int)      { gs[i], gs[j] = gs[j], gs[i] }
 func project2Tests() []test {
 	const label = "Project2"
 
-	sortGerms := cmp.FilterValues(func(x, y []*pb.Germ) bool {
-		ok1 := sort.IsSorted(germSorter(x))
-		ok2 := sort.IsSorted(germSorter(y))
-		return !ok1 || !ok2
-	}, cmp.Transformer("Sort", func(in []*pb.Germ) []*pb.Germ {
+	sortGerms := cmp.Transformer("Sort", func(in []*pb.Germ) []*pb.Germ {
 		out := append([]*pb.Germ(nil), in...) // Make copy
 		sort.Sort(germSorter(out))
 		return out
-	}))
+	})
 
 	equalDish := cmp.Comparer(func(x, y *ts.Dish) bool {
 		if x == nil || y == nil {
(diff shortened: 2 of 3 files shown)


cmp/options.go
```diff
@@ -183,9 +183,17 @@ func Ignore() Option {
...
-// If T and R are the same type, an additional filter must be applied to
-// act as the base case to prevent an infinite recursion applying the same
-// transform to itself (see the SortedSlice example).
+//
+// To help prevent infinite cycles recursively applying the same transform to
+// the output of itself (e.g., in the case where the input and output types
+// are the same), an implicit filter is added such that a transformer is
+// applicable only if that exact transformer is not already in the
+// tail of the Path since the last non-Transform step.
+// While this is sufficient for most usages of Transformers, this does not
```
jba: This caveat is unfortunate, and shouldn't be necessary. Also, describing the rule in terms of paths is technically precise but hard to understand intuitively. I would change the algorithm to better match mine (namely, each transformer can only be applied once per value per cmp.Equal call) and describe it in something like those terms.

dsnet: Are you concerned about how it is documented here? Or are you saying that the implementation is not equivalent to what you had in mind?

If the former, we can adjust the prose written.
If the later, what makes this current implementation different? I'm just trying to go for a minimal implementation so that we can test out ideas.

jba: Both.

I haven't looked at your actual code, but my thinking is that once you apply a transformer to a value in a given invocation of cmp.Equal, you shouldn't apply it again, whether or not it changes the path.

Basically, I'm thinking in terms of invocations of cmp.Equal, not path changes. Every invocation of cmp.Equal changes the path, but not every path change is the result of a call on cmp.Equal. I didn't realize until this discussion that you put things like type assertions and transforms in the path, and I don't totally get why you do, but in any case I'd ignore all the path steps that don't correlate with a call on Equal.

dsnet: It makes sense to have Transforms and TypeAssertions be part of the Path since they get printed. Otherwise you end up with weird situations when you know that the struct field type was an int, but the Path says its a string because there was a hidden transformation in between. For that reason, there is a one-to-one mapping of recursive calls to cmp.Equal and path steps.

The implementation of cmp.Equal more or less calls itself recursively exactly for each step:
* [`Indirect`](https://github.com/google/go-cmp/blob/d54e85569d47be6e6406055190a66430b6a905b7/cmp/compare.go#L251) for pointers
* [`TypeAssertion`](https://github.com/google/go-cmp/blob/d54e85569d47be6e6406055190a66430b6a905b7/cmp/compare.go#L264) for interfaces
* [`SliceIndex`](https://github.com/google/go-cmp/blob/d54e85569d47be6e6406055190a66430b6a905b7/cmp/compare.go#L470) for slices
* [`MapIndex`](https://github.com/google/go-cmp/blob/d54e85569d47be6e6406055190a66430b6a905b7/cmp/compare.go#L497) for maps
* [`StructField`](https://github.com/google/go-cmp/blob/d54e85569d47be6e6406055190a66430b6a905b7/cmp/compare.go#L540) for structs
* [`Transform`](https://github.com/google/go-cmp/blob/d54e85569d47be6e6406055190a66430b6a905b7/cmp/compare.go#L363) for transformations

The steps `Indirect`, `TypeAssertion`, `SliceIndex`, `MapIndex`, and `StructField` correspond directly with recursive calls dictated by the structure of the value themselves. Only `Transform` is special.

jba: OK, I see your point. I withdraw most of my criticism, though I still wish the doc would explain this more clearly.

PR #30, opened 2017-07-28
Fix Bug With Nil Stringers

mattdee123:
There was a bug in the way this package handles nil Stringer values.  This diff fixes the bug, and adds a previously failing test.


dsnet:
Thanks for the PR. Change LGTM, but you'll need to sign the CLA before I can submit this.

mattdee123:
I signed the thing -- does it normally take a while to propogate?

PR #32, opened 2017-08-01
Refactor option evaluation logic

dsnet:
The previous implementation of options had a single type "option"
that was used to represent either an Ignore, Comparer, or Transformer
and all of the filters relevant to each of them.

We refactor this logic by creating a new type to represent each
fundamental option and each filtering option. Construction of
filtered options is now as simple as wrapping the input option with
the appropriate filter type.

Evaluation of filters now takes a top-down approach where each of
the options now have a filter method and know how to filter themselves
and recursively call filter on the any sub-options.

This approach has the following advantages:
* It is faster because the same filter that was applied to multiple
options now only needs to execute once.
* Even though this is semantically identical, the implementation more
closely matches the documented algorithm in cmp.Equal.
* It allows for easier extension of the API to add new fundamental
and filtering option types.

We also add a String method to each of the fundamental and filter options.
This adds to the complexity of this commit, but helps users with debugging
since printing an option shows the full tree of everything applied.

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -54,11 +54,10 @@ var nothing = reflect.Value{}
 // If at least one Ignore exists in S, then the comparison is ignored.
 // If the number of Transformer and Comparer options in S is greater than one,
 // then Equal panics because it is ambiguous which option to use.
-// If S contains a single Transformer, then apply that transformer on the
-// current values and recursively call Equal on the transformed output values.
-// If S contains a single Comparer, then use that Comparer to determine whether
-// the current values are equal or not.
-// Otherwise, S is empty and evaluation proceeds to the next rule.
+// If S contains a single Transformer, then use that to transform the current
+// values and recursively call Equal on the output values.
+// If S contains a single Comparer, then use that to compare the current values.
+// Otherwise, evaluation proceeds to the next rule.
 //
 // • If the values have an Equal method of the form "(T) Equal(T) bool" or
 // "(T) Equal(I) bool" where T is assignable to I, then use the result of
@@ -119,8 +118,7 @@ type state struct {
 
 	// These fields, once set by processOption, will not change.
 	exporters map[reflect.Type]bool // Set of structs with unexported field visibility
-	optsIgn   []option              // List of all ignore options without value filters
-	opts      []option              // List of all other options
+	opts      Options               // List of all fundamental and filter options
 }
 
 func newState(opts []Option) *state {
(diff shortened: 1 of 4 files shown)


cmp/compare.go
```diff
@@ -283,90 +277,53 @@ func (s *state) compareAny(vx, vy reflect.Value) {
...
+	opt := evaluateOptions(s.opts, vx, vy, t, s.curPath, s.callTTBFunc)
+	if _, ok := opt.(ignore); !ok && (!vx.IsValid() || !vy.IsValid()) {
+		// If there were no FilterValues, we will not detect invalid inputs,
+		// so manually do a check and set invalid if necessary.
+		opt = invalid{}
 	}
-	return true
-}
 
-func (s *state) applyOption(vx, vy reflect.Value, t reflect.Type, opt option) {
```
neild: Can/should this function be replaced with a method on the Option interface?

```go
type Option interface {
  apply(vx, vy reflect.Value, t reflect.Type)
}
```

dsnet: It would need to be:
```go
apply(p Path, vx, vy reflect.Value, t reflect.Type) []Option
```

I don't mind plumbing `Path` all the time, but requiring every possible applicable option to allocate a slice seems very non-performant. For that reason, `evaluate` is method on `optionEvaluator` which only has a single slice to hold all that information.

dsnet: We might be able to avoid the performance hit by returning a single `Option` and relying on `Options` for the rare case where multiple are applicable. Let me look into this.

neild: You could plumb the `*state` instead of the `Path`.

This is all internal, so it doesn't matter *that* much, but it's weird to have an interface implemented by N types and then a big type switch on those types; isn't that precisely what interface methods are for?

cmp/options.go
```diff
@@ -35,79 +38,33 @@ type Option interface {
...
+			return ignore{} // Only ignore can short-circuit evaluation
+		case invalid:
+			out = invalid{} // Takes precedence over comparer or transformer
+		case *comparer, *transformer:
+			switch out.(type) {
+			case nil:
+				out = opt
+			case invalid:
+				// Keep invalid
+			case *comparer, *transformer:
```
dsnet: This is incorrect. I have to batch all *comparers and *transformers since it is still possible that a later `Ignore` will take precedence over all these conflicts.

dsnet: Fixed. I made it such that `filter` can return a `Options` only when there is a conflict.

dsnet:
I did your suggestion of removing the type switch adding each case as a method.

The reason I was confused was because I thought you were talking about adding that functionality to the `filter` method, which doesn't work.

I do like the extra bit of type safety this brings, where the return type of `filter` is `applicableOption`.

PR #43, opened 2017-09-27
Allow use of AllowUnexported on GopherJS

dsnet:
GopherJS does not have a complete implementation of unsafe, but it is
complete enough that it can do basic pointer arithmetic,
which is all we need to forcibly export an unexported field.

### DIFF
--- a/cmp/options.go
+++ b/cmp/options.go
@@ -348,7 +348,7 @@ func (cm comparer) String() string {
 // all unexported fields on specified struct types.
 func AllowUnexported(types ...interface{}) Option {
 	if !supportAllowUnexported {
-		panic("AllowUnexported is not supported on App Engine Classic or GopherJS")
+		panic("AllowUnexported is not supported on App Engine Classic")
 	}
 	m := make(map[reflect.Type]bool)
 	for _, typ := range types {
--- a/cmp/unsafe_panic.go
+++ b/cmp/unsafe_panic.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE.md file.
 
-// +build appengine js
+// +build appengine
 
 package cmp
 
--- a/cmp/unsafe_reflect.go
+++ b/cmp/unsafe_reflect.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE.md file.
 
-// +build !appengine,!js
+// +build !appengine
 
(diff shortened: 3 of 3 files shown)


dsnet:
For example: https://gopherjs.github.io/playground/#/WX6No0ltxL

dsnet:
I can't wait until we can delete the appengine tags too :)

shurcooL:
Interesting, I didn't know this was possible. :)

It _might_ be possible for me to update https://github.com/shurcooL/go-goon/blob/master/bypass/main.go based on this technique.

Also /cc @davecgh who might find this interesting.

shurcooL:
Also /cc @neelance.

In https://github.com/gopherjs/gopherjs/issues/170#issuecomment-72444436, he said:

> GopehrJS's support for `unsafe` is very restricted. It only supports some very special cases in a hacky way to make the `syscall` and `reflect` packages work. In general, `unsafe` can not be supported because that would require to emulate the memory characteristics of `gc`, including pointer arithmetics etc.

So I wonder if something has changed since then, or if this is working unintentionally (worse yet, maybe not working in all cases).

dsnet:
I believe this actually doesn't work properly. Printing the reflect.Value I retrieved from `unsafeRetrieveField` seems to indicate that it not quite what is expected. Sad. :(

shurcooL:
Okay, good to hear, because that's actually the expected outcome. GopherJS does not support `unsafe` (pointer arithmetic being a subset of that) this time, as its [compatibility table](https://github.com/gopherjs/gopherjs/blob/master/doc/packages.md) indicates.

shurcooL:
> Printing the reflect.Value I retrieved from `unsafeRetrieveField` seems to indicate that it not quite what is expected.

Can you share the snippet?

shurcooL:
I got 256 256, which matches what I get locally and on NaCL-based playground.

shurcooL:
@dsnet, I’d like to understand this better. You said:

> I believe this actually doesn't work properly. Printing the reflect.Value I retrieved from `unsafeRetrieveField` seems to indicate that it not quite what is expected.

Would you mind please sharing a snippet that lead you to that conclusion? Something where the output using GopherJS doesn’t match what’s expected. I tried and wasn’t able to so far.

dsnet:
Yep, sorry; got side-tracked by something else. Here's the snippet: https://gopherjs.github.io/playground/#/Jgf6BL8aYB

shurcooL:
Perfect, thank you!

PR #46, opened 2017-09-28
Use raw literal for string formatting

dsnet:
When formatting strings, avoid using strconv.Quote if the string looks
like it has already been escaped. Instead, use a raw string literal
by wrapping the string with '`' characters if possible.
For now, we still use strconv.Quote if the input string contains newlines
to maintain the property that Format outputs a single line.

Also, prefix strings obtained by the Stringer.String method with a 's'.
This allows users to more easily distinguish when an output really is
of type string or if the String method was used.

### DIFF
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -206,7 +206,7 @@ func comparerTests() []test {
 		label:    label,
 		x:        &struct{ R *bytes.Buffer }{new(bytes.Buffer)},
 		y:        &struct{ R *bytes.Buffer }{},
-		wantDiff: "root.R:\n\t-: \"\"\n\t+: <nil>\n",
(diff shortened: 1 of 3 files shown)


dsnet:
I couldn't think of anything better. The fundamental problem is that `Stringer.String` is an irreversible transform. I considered using `String() = ` as the prefix, but that looks overly verbose when formatting map keys or a slice of elements, each with a String method.

Regardless, I think some prefix is better than no prefix. The previous output was subtly lying to the user. We could avoid the `Stringer.String` altogether, but I believe there is value is having user defined outputs.

neild:
Yeah, I don't have a better alternative, and I agree that marking `Stringer` output is worthwhile. It's still going to cause confusion, though.

dsnet:
Submitting as is then. We can always adjust this in the future if someone has a better idea.

PR #48, opened 2017-10-02
Adjust heuristics for using raw literals

dsnet:
Rather than searching for escape sequences, a better measure of whether
to use raw literals or not is whether the raw literal form is shorter
than the quoted string encoding.
The assumption is that shorter string length is more readable,
which is a reasonable (except for non-printable characters).

### DIFF
--- a/cmp/internal/value/format.go
+++ b/cmp/internal/value/format.go
@@ -165,14 +165,22 @@ func formatAny(v reflect.Value, conf formatConfig, visited map[uintptr]bool) str
 }
 
 func formatString(s string) string {
-	// Avoid performing quote-escaping if the string is already escaped.
-	hasEscapes := strings.ContainsAny(s, `\`)
-	allPrintable := strings.IndexFunc(s, unicode.IsPrint) >= 0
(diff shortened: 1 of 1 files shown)


cmp/internal/value/format.go
```diff
@@ -165,14 +165,19 @@ func formatAny(v reflect.Value, conf formatConfig, visited map[uintptr]bool) str
...
 func formatString(s string) string {
-	// Avoid performing quote-escaping if the string is already escaped.
-	hasEscapes := strings.ContainsAny(s, `\`)
-	allPrintable := strings.IndexFunc(s, unicode.IsPrint) >= 0
-	rawAllowed := !strings.ContainsAny(s, "`\n")
-	if hasEscapes && allPrintable && rawAllowed {
-		return "`" + s + "`"
+	// Use raw string literals if it is shorter than the quoted string.
+	qs := strconv.Quote(s)
+	if len(qs) > 1+len(s)+1 {
```
neild: Flip this:

```go
if len(qs) <= 1+len(s)+1 {
  return qs
}
// etc.
```

PR #49, opened 2017-10-05
Explicitly convert assignable nil interface values

dsnet:
The reflect package has a bug (golang/go#22143).
Let R and T both be interface types and R be assignable to T.
Let F be a function that takes T as its argument.

The Go language lets you directly call F with a nil T or a nil R
(since R is assignable to T) without any special handling.

Currently, reflect allows you to call F with a nil T,
but panics when calling F with a nil R. To avoid this panic,
we explicitly check for the condition where a value:
	* is an interface type
	* is nil
	* is not exactly same as type T
and re-create the value as a nil interface of type T.

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -322,6 +322,7 @@ func (s *state) tryMethod(vx, vy reflect.Value, t reflect.Type) bool {
 }
 
 func (s *state) callTRFunc(f, v reflect.Value) reflect.Value {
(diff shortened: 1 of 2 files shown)


shurcooL:
> The reflect package has a bug (golang/go#22143).

Fun observation, the GopherJS implementation of `reflect` package (a custom one) does not have this bug. https://gopherjs.github.io/playground/#/bRzJ0dQ_0W prints 4 `true`s, as expected.

PR #50, opened 2017-10-25
Bump minimum version to Go1.8

dsnet:
Go1.8 went GA on AppEngine.
The lack of Go1.8 support was the primary reason for much
of these backwards compatibility hacks.

Bumping to Go1.8 still ensures that we're supporting at least
the latest 2 versions of Go, which are Go1.8 and Go1.9.


jba:
Not yet, please. The cloud libraries may still need to support 1.6.
/cc @zombiezen

dsnet:
I'll be holding off on this until Jan 2nd, 2018.

jba:
Please don't merge. We're going to have to support 1.6 for a while longer.

dsnet:
Do I still have to hold off on this?

According to https://cloud.google.com/appengine/docs/standard/go/release-notes, Go1.8 has been GA since October 25th, 2017.

jba:
Until AppEngine is off of 1.6, we need to support it.

golang.org/x/oauth2 still works with 1.6 (since we depend on it).

dsnet:
(If I understood correctly) In a meeting we had some time ago, my understanding was to only support the last 2 versions of Go on AppEngine and that anything older would be the responsibility of the user to vendor their dependencies. Now that AppEngine has both Go1.8 and Go1.9, doesn't that remove our responsibility to keep supporting Go1.6?

If we do have keep Go1.6 support, when is that going to be removed from AppEngine? (Tangent) I should also note, that starting on April 30th, we're going to merge the `dev` branch of protobufs, which will effectively raise the minimum version to Go1.9.

dsnet:
With the release of [Go1.11 on AppEngine](https://www.google.com/url?q=https://cloud.google.com/blog/products/application-development/go-1-11-is-now-available-on-app-engine&sa=D&source=hangouts&ust=1539804678757000&usg=AFQjCNEWsIdkDblXlL6edZ7TaQNqml0t_w), I'd like to revisit dropping support for anything lower than Go1.8. I'd like to make the equivalent change to protocol buffers as well.

Anyone objections to dropping support?

\cc @sbuss @jba @neild @zombiezen

jba:
Wait until Nov. 1.

dsnet:
Times up!

PR #53, opened 2017-11-06
Add IgnoreUnsetFields helper to cmpopts

dsnet:
IgnoreUnsetFields is useful for performing the equivalent of a "partial"
equality, where only fields based on some template struct are compared,
any everything else is ignored.

This is useful for forward-compatible comparison where the user wants
to ensure struct fields added in the future do not get compared.

### DIFF
--- a/cmp/cmpopts/ignore.go
+++ b/cmp/cmpopts/ignore.go
@@ -7,23 +7,61 @@ package cmpopts
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"unicode"
 	"unicode/utf8"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/internal/value"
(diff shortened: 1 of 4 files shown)


jba:
This could be confusing.
```
IgnoreUnsetFields(MyStruct{A: 1, B:0}
```
B is ignored even though I "set" it. And that will almost never get caught, because it only increases the number of tests that pass.

At a minimum, rename to `IgnoreZeroFields`.

dsnet:
I'm revoking this PR. I don't like the fact that I this API forces me to think about what is the non-zero value of a type. A better API would probably be `IgnoreFieldsExcept(typ interface{}, names ...string)` which is similar to `IgnoreFields`, but ignores all fields not in the given list. I'm not going to add the new option until I see more cases where it would have been useful.

PR #56, opened 2017-12-01
Add Path.Index helper method

dsnet:
An analysis of user code shows that Path.Last was not sufficient and
that what users really wanted was an easy want to do negative indexing
without needing to constantly check for out-of-bounds.

Thus, add a generic Index method that does not panic on out-of-bounds
accesses and supports negative indexing.


dsnet:
I regret adding the Last method and wished I went with the Index method the first time.

cmp/path.go
```diff
@@ -94,10 +94,19 @@ func (pa *Path) pop() {
 // Last returns the last PathStep in the Path.
 // If the path is empty, this returns a non-nil PathStep that reports a nil Type.
 func (pa Path) Last() PathStep {
-	if len(pa) > 0 {
-		return pa[len(pa)-1]
+	return pa.Index(-1)
+}
+
+// Index returns the ith step in the Path and supports negative indexing.
```
neild: Pedantic: While anyone familiar with Python knows what negative indexing is, if you want to be precise you should define a negative index.

dsnet: Done.

PR #58, opened 2017-12-08
Support AllowUnexported on gopherjs

dsnet:
Use the js package to directly mutate the RO bits in reflect.Value.
It assumes that reflect.Value has a field named flag and that the constants have not changed.

### DIFF
--- a/cmp/unsafe_gopherjs.go
+++ b/cmp/unsafe_gopherjs.go
@@ -1,0 +1,29 @@
+// Copyright 2017, The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE.md file.
+
+// +build js
+
+package cmp
+
+import (
+	"reflect"
+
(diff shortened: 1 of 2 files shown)


dsnet:
\cc @shurcooL

What would it take to setup Travis to run gopherjs? I'm curious how much stuff breaks. I don't the tests pass cleanly.

shurcooL:
The `master` version of GopherJS requires Go 1.9. Aside from that, it should work with just `go get -u github.com/gopherjs/gopherjs` as per https://github.com/gopherjs/gopherjs#installation-and-usage.

However, to be able to use `gopherjs test`, depending on whether the tests do file IO, etc., you might need to install the `node-syscall` module (as well as `node` of course), as per https://github.com/gopherjs/gopherjs#gopherjs-run-gopherjs-test.

I've never tried to run GopherJS tests in Travis; I've always relied on `go test` so far, so I don't have prior experience with this. But I'm happy to help investigate.

For my benefit, can you summarize what you're trying to do?

PR #62, opened 2017-12-21
A regression test for a brittle Equal method.

filmil:
This seems to me to demonstrate that the contract for the Equal method should
either be precisely specified, or that go-cmp should be fixed to ensure that
it behaves sensibly when encountered with nils.

### DIFF
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -48,6 +48,7 @@ func TestDiff(t *testing.T) {
 	tests = append(tests, project2Tests()...)
 	tests = append(tests, project3Tests()...)
 	tests = append(tests, project4Tests()...)
+	tests = append(tests, nilEqualityTests()...)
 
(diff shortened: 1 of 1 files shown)


dsnet:
Anytime you indirect a pointer you should either:
* Check whether the pointer is nil
* Verify that the code leading up to that point guarantees that the pointer is set
* Or decide that a panic is exactly what you want.

For an Equal method there is a sensible definition of what should happen:
```go
func (x *foo) Equal(y *foo) bool {
	if x == nil || y == nil {
		return x == nil && y == nil
	}
	return x.value == y.value
}
```

Thus, I argue that this is a bug in the definition of Equal.

Alternatively, the Equal method could be defined on the non-pointer receiver version. The `time.Time` type uses this pattern.

dsnet:
Closing this as it seems this was demonstrate the problem with the associated issue.

PR #63, opened 2017-12-21
Document that Equal is called even if x or y is nil

dsnet:


filmil:
LGTM

cmp/compare.go
```diff
@@ -61,8 +61,8 @@ var nothing = reflect.Value{}
 //
 // • If the values have an Equal method of the form "(T) Equal(T) bool" or
 // "(T) Equal(I) bool" where T is assignable to I, then use the result of
-// x.Equal(y). Otherwise, no such method exists and evaluation proceeds to
-// the next rule.
+// x.Equal(y) even if x or y are nil.
```
jba: s/are/is/

PR #65, opened 2018-01-03
Report with primitive types in diffs

### LINKED ISSUE #64: diff output with multiple differences is unhelpful
I ran across this recently when using cmp to test differences in unmarshal results.
One unmarshaler was unmarshaling numbers as floats, the other as ints, but the
Diff result didn't make it at all clear what the difference was.

Example code:

	package main
	
	import (
		"fmt"
	
	)
	
	func main() {
		x := []interface{}{map[string]interface{}{"avg": 0.278, "hr": 65, "name": "Mark McGwire"}, map[string]interface{}{"avg": 0.288, "hr": 63, "name": "Sammy Sosa"}}
		y := []interface{}{map[string]interface{}{"avg": 0.278, "hr": 65.0, "name": "Mark McGwire"}, map[string]interface{}{"avg": 0.288, "hr": 63.0, "name": "Sammy Sosa"}}
	
		diff := cmp.Diff(x, y)
		fmt.Printf("diff: %s\n", diff)
	}

This prints:

	diff: root:
		-: []interface {}{map[string]interface {}{"avg": 0.278, "hr": 65, "name": "Mark McGwire"}, map[string]interface {}{"avg": 0.288, "hr": 63, "name": "Sammy Sosa"}}
		+: []interface {}{map[string]interface {}{"avg": 0.278, "hr": 65, "name": "Mark McGwire"}, map[string]interface {}{"avg": 0.288, "hr": 63, "name": "Sammy Sosa"}}

Note that both lines are identical.

dsnet:
When the formatted strings are identical, the reporter tries to fallback
on more exact formatting. This fallback should include the types for
primitive types since the difference can simply be between
an int and an uint.

Fixes #64

### DIFF
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -453,6 +453,17 @@ root:
 				return x == nil && y == nil
 			}, cmp.Ignore()),
 		},
+	}, {
+		label: label,
+		x:     []interface{}{map[string]interface{}{"avg": 0.278, "hr": 65, "name": "Mark McGwire"}, map[string]interface{}{"avg": 0.288, "hr": 63, "name": "Sammy Sosa"}},
+		y:     []interface{}{map[string]interface{}{"avg": 0.278, "hr": 65.0, "name": "Mark McGwire"}, map[string]interface{}{"avg": 0.288, "hr": 63.0, "name": "Sammy Sosa"}},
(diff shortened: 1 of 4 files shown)


rogpeppe:
Nice. Thanks very much!

PR #68, opened 2018-02-02
Support purego build tag

dsnet:
The proposal in golang/go#23172 was accepted. The "purego" build tag
is intended to be a community agreed upon soft-signal to indicate the
forbidden use of unsafe, assembly, or cgo.

A change in the future will remove special-casing the appengine and js tags
once the related toolchains support purego (possibly after some bake-in period).


dsnet:
> The text should be updated to reflect the possibility of seeing that panic due to using "purego" build tag.

Done.

Issue #67, opened 2018-02-01
proposal: use generic less function when cmpopts.SortSlices argument is nil


rogpeppe:
It's a bit of a pain writing a comparison function when we want slices to compare equal regardless of order. How about making SortSlices use a generic less function when it's argument is nil, so that in the common case we can just use SortSlices(nil) to check slice contents regardless of order and type?

The generic less function might be something like this, perhaps: https://play.golang.org/p/NQ1u36jKsi-

dsnet:
For Bool, Int, Int8, Int16, Int32, Int64, Uint, Uint8, Uint16, Uint32, Uint64, Uintptr, the comparison is obvious.
For Float32 and Float64, we [have to handle NaN](https://github.com/golang/go/blob/3caa02f/src/sort/sort.go#L283), but that's trivial.
For Complex64 and Complex128, we could sort them as if they were a struct{Real, Imag float64}.
For String, Slice, Array, and Struct, we could use lexicographical sorting of their elements/fields.
For Ptr and Interface, we could sort the object being pointed to; treat nil as less-than non-nil; if types are different, sort by type name (yuck, but deterministic).
For Chan, Func, UnsafePointer, the only sensible way to sort these is by pointer. I'm fairly opposed to resorting to pointer comparison since it is highly runtime dependent. It might be better to panic on such comparisons.
For Maps, I don't have any idea how to sort that.

My overall thoughts/questions:
* What's the right behavior with regard to Chan, Func, UnsafePointer, and Maps?
* For structs, what's the right policy with regard to unexported fields? E.g., Suppose a struct has an unexported field at the top that is a sync.Mutex, it would be mistake to have any sorting behavior be dependent on the unexported fields of the Mutex.

\cc @bcmills who originally pushed for `cmp` to include first-class support for ordering. (`cmp` was originally called `eq`, but renamed to allow for the possibility to ordering).

neild:
Panic on chan and func. I'd just convert UnsafePointer to a uintptr; it's runtime dependent but so are most uintptrs.

Defining an ordering for maps is straightforward, but I don't see the use case. Maybe panic there as well. If you want an ordering, then: Take the union of all keys in both maps, sort it, compare the value for each key, first inequality defines the ordering. An absent key is less than a present one.

jba:
[...]

PR #71, opened 2018-02-13
cmp/cmpopts: implement generic less function

rogpeppe:
Implement GenericLess, as suggested in #67 and make SortSlices use it.
It should work for all comparable values.

It doesn't seem worth making SortMaps use GenericLess, as the main reason for SortMaps seems to be to enable comparing of values in ways that aren't related to the Go equality comparison (e.g. time.Time).

### DIFF
--- a/cmp/cmpopts/sort.go
+++ b/cmp/cmpopts/sort.go
@@ -6,14 +6,19 @@ package cmpopts
 
 import (
 	"fmt"
+	"math"
 	"reflect"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/internal/function"
 )
 
 // SortSlices returns a Transformer option that sorts all []V.
-// The less function must be of the form "func(T, T) bool" which is used to
+//
+// If less is nil, GenericLess will be used, which works on any comparable element type.
+//
+// Otherwise, the less function must be of the form "func(T, T) bool" which is used to
 // sort any slice with element type V that is assignable to T.
 //
 // The less function must be:
@@ -26,6 +31,9 @@ import (
 //
 // SortSlices can be used in conjunction with EquateEmpty.
 func SortSlices(less interface{}) cmp.Option {
+	if less == nil {
+		less = GenericLess
+	}
 	vf := reflect.ValueOf(less)
 	if !function.IsType(vf.Type(), function.Less) || vf.IsNil() {
 		panic(fmt.Sprintf("invalid less function: %T", less))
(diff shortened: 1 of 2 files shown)


cmp/cmpopts/sort.go
```diff
@@ -144,3 +155,126 @@ func (ms mapSorter) less(v reflect.Value, i, j int) bool {
...
+
+// GenericLess reports whether x is less than y, where both x and y must
+// be comparable types. For basic types, the values are compared by
+// value (pointer values are compared by pointer value, so ordering is
+// runtime-dependent in this case); for composite types, values are
+// compared lexicographically by component. Float and complex NaN values
+// will be compared as equal to avoid potential panics when sorting.
+//
+// GenericLess panics if provided with non-comparable values.
+func GenericLess(x, y interface{}) bool {
```
dsnet: I have concerns that will reduce the composability of options.

The fact that this is `interface{}` implies that `cmpopts.SortSlices` will be applicable on any `[]T`. Thus, the following will panic:
```go
// Hexdump transformation for better diffing.
hexdumper := cmp.Transformer(func (v []byte) []string {
    return strings.Split(hex.Dump(v), "\n")
})
cmp.Equal(x, y, cmpopts.SortSlices(nil), hexdumper)
```
This panics because it is ambiguous whether to use `SortSlices` or `hexdumper` for a value of type `[]byte`.

cmp/cmpopts/sort.go
```diff
@@ -144,3 +155,126 @@ func (ms mapSorter) less(v reflect.Value, i, j int) bool {
...
+		return 0
+	case reflect.Ptr, reflect.Chan, reflect.UnsafePointer:
+		xi, yi := xv.Pointer(), yv.Pointer()
+		switch {
+		case xi < yi:
+			return -1
+		case xi > yi:
+			return 1
+		}
+		return 0
```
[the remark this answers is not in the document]
rogpeppe: Yes, I agree that this behavior would be useful in some cases. But if we allow sorting by pointer, then we'd have to deal with circular references, which would complicate the implementation significantly, and raise some awkward questions too.

I think that for this kind of case, it's still OK to require users to write their own comparison function as they do now. GenericLess doesn't claim to cover all possible cases, but hopefully a significantly useful subset, and one which is readily understandable to those that understand Go, because the set of types that can be compared is exactly those of Go.

That's why I chose that particular semantic. There are always going to be cases which slip over the line, but I still think it's worth having *some* implementation of GenericLess for the large set of cases which fall the right side of the line.

cmp/cmpopts/sort.go
rogpeppe: Yes, that's indeed the case, but perhaps it's made a little more palatable by the fact that the panic message says immediately what the issue is and suggests how to solve it.

	cmp.Equal(x, y, cmp.FilterPath(func(p cmp.Path) bool {
		return p.Last().Type() != reflect.TypeOf([]byte(nil))
	}, cmpopts.SortSlices(nil)), hexdumper)

or perhaps:

	cmp.Equal(x, x, cmpopts.SortSlices(func(s, t string) bool {
		return cmpopts.GenericLess(s, t)
	}), hexdumper)

That is, the options are still composable with a little more effort.

rogpeppe:
I guess this isn't going to happen. For anyone watching, here's a hacky solution we came up with elsewhere: use:

    cmpopts.SortSlices(func(x, y interface{}) bool {

         return pretty.Sprint(x) < pretty.Sprint(y)
    })

dubious but short and effective.

PR #76, opened 2018-03-09
Export FilterField

### LINKED ISSUE #75: question: convenient FilterPath functions
Hello, I've been[ using `go-cmp`](https://godoc.org/github.com/gotestyourself/gotestyourself/assert#DeepEqual) for a bit now and I've found that the two most common cases for needing an `Option` are:
1. `AllowUnexported`/`IgnoreUnexported` - which is already easy to use
2. A `FilterPath` for fields that have values set from `time.Now()`  (`time.Time`, or `time.Duration`)

It wasn't immediately obvious to me how to create a correct `FilterPath` filter function. After a quick scan of the godoc my first approach was something like this:

```go
func(path gocmp.Path) bool {		
    return path.Last().String() == ".Elapsed"		
}
```
This worked, but it seems to me like it could potentially match fields that I don't want it to match if I used it with a large nested structure where multiple types had an `Elapsed` field.

Digging into the godoc further I noticed that a bunch of the types embedded `PathStep`, which made me realize that a more correct way of using `Path` would probably be to iterate over the steps, type assert to the expected type, and compare to `Name()` or `Key()`.

My next iteration (https://github.com/gotestyourself/gotestyourself/pull/62) looks something like this: 

```go
func fieldPathWithTypes(pathspec string) func(gocmp.Path) bool {
	return func(path gocmp.Path) bool {
		spec := strings.Split(pathspec, ".")
		index := 0

		for _, step := range path {
			fieldStep, ok := step.(gocmp.StructField)
			if !ok {
				continue
			}
			if index >= len(spec) || spec[index] != fieldStep.Name() {
				return false
			}
			index++
		}
		return index == len(spec)
	}
}
```

Which could be made a lot simpler if it's expected that `PathStep.String()` is stable and wont change between minor versions:

```go
func fieldPathSimple(pathspec string) func(gocmp.Path) bool {
	return func(path gocmp.Path) bool {
		return path.String() == pathspec
	}
}
```

The same idea could be applied using `GoString()` if that is considered stable. The idea behind both of these functions is that it's much easier for a reader to understand a dotted path string, than it is to read a long function with a bunch of type assertions.

Questions:
1. What is the expected/"best practice" way of using `PathStep`? Is it similar to `fieldPathWithTypes` ?
2. Would you be interested in a PR that adds a convenience function for building the `PathFilter` filter function from a simple string? This could be something like `fieldPathWithTypes` or it could be more verbose (to match against something like `GoString()`. I believe the only type that can't be easily translated from a string representation would be `Transform`.
3. Would you be interested in a PR that adds one or more examples for using `FilterPath` (based on feedback from these other questions) ?

dnephin:
Closes #75

Opening for discussion. I guess there could be more tests cases as well, but all the code is effectively covered by other cases.

### DIFF
--- a/cmp/cmpopts/struct_filter.go
+++ b/cmp/cmpopts/struct_filter.go
@@ -12,17 +12,13 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
-// filterField returns a new Option where opt is only evaluated on paths that
+// FilterField returns a new Option where opt is only evaluated on paths that
 // include a specific exported field on a single struct type.
 // The struct type is specified by passing in a value of that type.
 //
 // The name may be a dot-delimited string (e.g., "Foo.Bar") to select a
 // specific sub-field that is embedded or nested within the parent struct.
-func filterField(typ interface{}, name string, opt cmp.Option) cmp.Option {
-	// TODO: This is currently unexported over concerns of how helper filters
-	// can be composed together easily.
-	// TODO: Add tests for FilterField.
-
+func FilterField(typ interface{}, name string, opt cmp.Option) cmp.Option {
 	sf := newStructFilter(typ, name)
 	return cmp.FilterPath(sf.filter, opt)
 }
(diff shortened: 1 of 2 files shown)


dnephin:
I'll close this until the discussion in #75 results in something actionable.

Issue #74, opened 2018-03-01
proposal: support cyclic data structures


posener:
Hi
The following program crashes on an infinite recursion:

```go
package main

import "github.com/google/go-cmp/cmp"

type Node struct {
	Next *Node
}

func main() {
	a := &Node{}
	a.Next = a

	cmp.Equal(a, a)
}
```

Fixing this is needed to use this package in [testify](https://github.com/stretchr/testify).

I saw that there is already a [TODO comment](https://github.com/google/go-cmp/blob/master/cmp/compare.go#L183).

dsnet:
Properly supporting cyclic data structures is very challenging. Even though I added that TODO, over time I've become convinced that `cmp` shouldn't support cyclic structures by default in order to avoid further complicating its logic.

I'm aware that `reflect.DeepEqual` supports cyclic structures, but their definition of equality is strange and probably not what the user expects. Consider [this example](https://play.golang.org/p/ku21y8f2lDq). The graphs have clearly different layouts, yet they are all considered the same.

I should note that some cyclic structures are only cyclic because the structure is conceptually a tree, but every node has a "weak" reference to the root. In those situations, you can use an Ignore option to prevent `cmp` from traversing back to the root.

If someone has a good design for supporting graph equality (and fits in well with the ecosystem of `cmp.Option`), I'll consider it. For now, I have no intention of adding this functionality.

posener:
Thanks for the detailed explanation! :+1: 

What about this sort of solution?
Add a `memory` variable to the `state` struct, that will be a set of a pair of addresses: those addresses are addresses of variables that their comparisons was already started.
We can add the addresses to this set in the `compareAny` switch, in `reflect.Struct` case.
Then, in the `compareAny` switch, `reflect.Ptr` case, we can add a test that if the pointers point to addresses that are already in the set, we skip the recursive call.

dsnet:
In my opinion, it is not sufficient to simply just check whether we have visited a pointer before. If you read the rules of `reflect.DeepEqual`, that is exactly what they do:
> As DeepEqual traverses the data values it may find a cycle. The second and subsequent times that DeepEqual compares two pointer values that have been compared before, it treats the values as equal rather than examining the values to which they point.

This approach is why `reflect.DeepEqual` has strange results where graphs of clearly different structures are considered equal.

I think you would need to do something like this:
* After descending into a pointer, record the current `cmp.Path` and the pointer address.
[...]

PR #78, opened 2018-03-17
Add DetectCycles to detect cycles in compared structs

posener:
The current comparison fails on stack overflow when given a cyclic struct
(such as a cyclic linked list for example).

This fix adds a new option: DecetctCycles, which finds cycles by adding information
to the path stack, when iterating over pointers.

When a cycle is checked (every time a pointer kind is compared) the path stack
is iterated to find cycles. The comparison of cycles is by they length.

Fixes: #74

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -119,6 +119,10 @@ type state struct {
 	// These fields, once set by processOption, will not change.
 	exporters map[reflect.Type]bool // Set of structs with unexported field visibility
 	opts      Options               // List of all fundamental and filter options
+
+	// detectCycles is set by the DetectCycles option, and it indicates to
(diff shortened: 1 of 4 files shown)


posener:
@dsnet , I can improve this fix.
Instead of saving `xAddr`, and `yAddr` in the path stack, I can add a map to the `cycleDetector` struct that will map the pointer address to the stack length. - then the algorithm becomes O(1) time complexity.

If I make this improve, we can make the cycle detection enabled by default, cause it won't increase the time complexity of the algorithm, and the space complexity will be in the same order it was before.

What do you say?

posener:
I think that #79 is a better solution, but I'm leaving this one open for your decision.

dsnet:
I'll comment on #79.

PR #81, opened 2018-03-20
Add pointer comparison in compareAny

### LINKED ISSUE #80: Comparing reflect types behaves not as expected
When comparing two equal types, the response is false, for example, the following code print `false`:

```go
t1 := reflect.TypeOf(1)
t2 := reflect.TypeOf(2)
fmt.Println(cmp.Equal(t1, t2, deepAllowUnexported(t1)))
```
[See full code](https://github.com/google/go-cmp/files/1830968/main.go.txt)

Just to clarify, if the diff is being printed, this is the reason that those two types are not the same:
```
{*reflect.rtype}.alg.hash:
	-: (func(unsafe.Pointer, uintptr) uintptr)(0x451530)
	+: (func(unsafe.Pointer, uintptr) uintptr)(0x451530)
{*reflect.rtype}.alg.equal:
	-: (func(unsafe.Pointer, unsafe.Pointer) bool)(0x401a50)
	+: (func(unsafe.Pointer, unsafe.Pointer) bool)(0x401a50)
```

This is being solved in the standard `refelect.DeepEqual` [when comparing two pointers](https://github.com/golang/go/blob/master/src/reflect/deepequal.go#L95-L97).

If the `compareAny` will do the same trick, the above code will print `true` as expected. It also does not break any of the tests:

```diff
diff --git a/cmp/compare.go b/cmp/compare.go
index 1260603..8fbbaa9 100644
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -239,6 +239,12 @@ func (s *state) compareAny(vx, vy reflect.Value) {
                        s.report(vx.IsNil() && vy.IsNil(), vx, vy)
                        return
                }
+
+               if vx.Pointer() == vy.Pointer() {
+                       s.report(true, vx, vy)
+                       return
+               }
+
```

I am not sure why this comparison is not done in the `go-cmp` implementation, I can't see any harm in doing it, it can also saves running time.

posener:
When comparing pointers in compareAny, if the pointers are pointing on the
same object we report that they are equal.

Fixes #80

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -239,6 +239,13 @@ func (s *state) compareAny(vx, vy reflect.Value) {
 			s.report(vx.IsNil() && vy.IsNil(), vx, vy)
 			return
 		}
+
+		// If pointers are pointing on the same thing, we can report them as equal
+		if vx.Pointer() == vy.Pointer() {
+			s.report(true, vx, vy)
+			return
+		}
+
 		s.curPath.push(&indirect{pathStep{t.Elem()}})
 		defer s.curPath.pop()
 		s.compareAny(vx.Elem(), vy.Elem())
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -464,6 +464,10 @@ root[0]["hr"]:
 root[1]["hr"]:
 	-: int(63)
 	+: float64(63)`,
+	}, {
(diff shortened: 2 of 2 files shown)


posener:
As explained in #80, this is not necessary.

PR #85, opened 2018-03-27
Add support for comparing graphs

### LINKED ISSUE #74: proposal: support cyclic data structures

dsnet:
Previously, trying to call Equal on a graph would result in a stack-overflow
due to infinite recursion traversing cycles on a graph.
While a vast majority of Go values are trees or acyclic graphs, there exist
a small number of cases where graph equality is required.

As such, we add cycle detection to Equal and define what it means for two graphs
to be equal. Contrary to reflect.DeepEqual, which declares two graphs to be
equal so long any cycle were encountered, we require two graphs to have
equivalent graph structures.

Mathematically speaking, a graph G is a tuple (V, E) consisting of the set of
vertices and edges in that graph. Graphs G1 and G2 are equal if V1 == V2,
E1 == E2, and both have the same root vertex (entry point into the graph).
When traversing G1 and G2, we remember a stack of previously visited edges
ES1 and ES2. If the current edge e1 is in ES1 or e2 is in ES2, then we know that
a cycle exists. The graphs have the same structure when the previously
encountered edge ep1 and ep2 were encountered together.
Note that edges and vertices unreachable from the root vertex are ignored.

Appreciation goes to Eyal Posener (@posener), who proposed a different
(but semantically equivalent) approach in #79, which served as inspiration.

Fixes #74

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -80,6 +80,11 @@ import (
 // Pointers and interfaces are equal if they are both nil or both non-nil,
 // where they have the same underlying concrete type and recursively
 // calling Equal on the underlying values reports equal.
+//
+// Before recursing into a pointer, slice element, or map, the current path
+// is checked to detect whether the address has already been visited.
+// If there is a cycle, then the pointed at values are considered equal
+// only if both addresses were previously visited in the same path step.
 func Equal(x, y interface{}, opts ...Option) bool {
 	vx := reflect.ValueOf(x)
 	vy := reflect.ValueOf(y)
@@ -137,6 +142,7 @@ type state struct {
 	// Calling statelessCompare must not result in observable changes to these.
 	result    diff.Result // The current result of comparison
 	curPath   Path        // The current path in the value tree
+	curPtrs   pointerPath // The current set of visited pointers
 	reporters []reporter  // Optional reporters
 
 	// recChecker checks for infinite cycles applying the same set of
@@ -155,6 +161,7 @@ type state struct {
 func newState(opts []Option) *state {
 	// Always ensure a validator option exists to validate the inputs.
 	s := &state{opts: Options{validator{}}}
+	s.curPtrs.Init()
 	s.processOption(Options(opts))
 	return s
 }
@@ -187,9 +194,9 @@ func (s *state) processOption(opt Option) {
 // This function is stateless in that it does not alter the current result,
 // or output to any registered reporters.
 func (s *state) statelessCompare(step PathStep) diff.Result {
-	// We do not save and restore the curPath because all of the compareX
-	// methods should properly push and pop from the path.
-	// It is an implementation bug if the contents of curPath differs from
+	// We do not save and restore curPath and curPtrs because all of the
+	// compareX methods should properly push and pop from them.
+	// It is an implementation bug if the contents of the paths differ from
 	// when calling this function to when returning from it.
 
 	oldResult, oldReporters := s.result, s.reporters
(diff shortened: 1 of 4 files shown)


dsnet:
@neild for review. I still need to add a lot more unit tests.

cmp/compare.go
```diff
@@ -399,21 +430,34 @@ func (s *state) compareArray(vx, vy reflect.Value, t reflect.Type) {
...
-	// Compute an edit-script for slices vx and vy.
-	es := diff.Difference(vx.Len(), vy.Len(), func(ix, iy int) diff.Result {
-		step.xkey, step.ykey = ix, iy
-		return s.statelessCompare(vx.Index(ix), vy.Index(iy))
-	})
-
-	// Report the entire slice as is if the arrays are of primitive kind,
-	// and the arrays are different enough.
+	// Checking element pointers is necessary for slices of non-primitives,
+	// as they may contain slices that sub-slices some parent slice.
```
neild: typo: s/sub-slices/sub-slice/

cmp/path.go
```diff
@@ -291,6 +293,72 @@ var (
...
+// In Go, cycles can occur via pointers, slices, and maps.
+//
+// The pointerPath uses a map to represent a stack; where descension into a
+// pointer pushes the address onto the stack, and ascension from a pointer
+// pops the address from the stack. Thus, when traversing into a pointer from
+// reflect.Ptr, reflect.Slice element, or reflect.Map, we can detect cycles
+// by checking whether the pointer has already been visited. The cycle detection
+// uses a seperate stack for the x and y values.
+//
+// Suppose a cycle were detected, we need to determine whether the two pointers
```
neild: "If a cycle is detected we need to..."

cmp/path.go
```diff
@@ -291,6 +293,72 @@ var (
...
+	my map[pointer.P]pointer.P
+}
+
+func (p *pointerPath) Init() {
+	p.mx = make(map[pointer.P]pointer.P)
+	p.my = make(map[pointer.P]pointer.P)
+}
+
+// Push indicates intent to descend into pointers vx and vy where
+// visited reports whether thay have been seen before. If visited before,
```
neild: reports whether either has been seen before

cmp/compare.go
```diff
@@ -267,6 +281,23 @@ func (s *state) compareAny(vx, vy reflect.Value) {
...
+		// For example v[:0] and v[:1] have the same slice pointer,
+		// but they are clearly different values. Using the slice pointer
+		// violates the assumption that equal pointers implies equal values.
+		/*
+			if eq, visited := s.curPtrs.Push(vx, vy); visited {
+				s.report(eq, vx, vy)
+				return
+			}
+			defer s.curPtrs.Pop(vx, vy)
+		*/
```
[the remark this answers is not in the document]

cmp/compare.go

cmp/path.go
dsnet: Done.

cmp/path.go
dsnet: Done.

posener:
@dsnet , it have been a while :smile: , any intentions to merge this?

dsnet:
Apologies for the massive delay. This is next on my TODO list in regards to `cmp` now that the reporter logic has been completely re-written.

LasTshaMAN:
Any updates on this ?

PR #91, opened 2018-07-09
Swap order of want and got in Diff example

zombiezen:
I suppose this is debatable, but I tend to think about diffs as "from golden output, what is different?" to see that the system under test added something it shouldn't have or removed something it shouldn't have.

### DIFF
--- a/cmp/example_test.go
+++ b/cmp/example_test.go
@@ -92,12 +92,12 @@ func ExampleDiff_testing() {
 	for _, test := range tests {
 		AddCrew(test.before, test.name, test.title)
 		if diff := cmp.Diff(test.before, test.after); diff != "" {
-			t.Errorf("%s: after AddCrew, manifest differs: (-got +want)\n%s", test.desc, diff)
+			t.Errorf("%s: after AddCrew, manifest differs: (-want +got)\n%s", test.desc, diff)
 		}
 	}
 
 	// Output:
-	// add to empty: after AddCrew, manifest differs: (-got +want)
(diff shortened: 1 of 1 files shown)


dsnet:
A quick search across tests in Google shows around 40% usage of `-got +want` and 60% usage of `-want +got`. I consider myself in the `-got +want` camp, but can't deny what the slight majority prefers.

cmp/example_test.go
```diff
@@ -92,28 +92,28 @@ func ExampleDiff_testing() {
 	for _, test := range tests {
 		AddCrew(test.before, test.name, test.title)
 		if diff := cmp.Diff(test.before, test.after); diff != "" {
```
dsnet: Perhaps:
`s/before/want/g`
`s/after/got/g`
to make it more consistent with the documented string.

zombiezen: This example is quite confusing. I tried to make the change you suggested, but it becomes quite weird to mutate a field called `got` to have the result. WDYT?

dsnet:
FYI. I the time in-between, it seems that usage has shifted to be 20% being "-got +want" and 80% being "-want +got".

cmp/example_test.go
dsnet: We can leave it as is.

dsnet:
Is there more to do, or shall I submit?

zombiezen:
I don't have anything more to add. Feel free to carry submit/carry forward.

PR #94, opened 2018-09-11
go.mod: add module file

dsnet:
Go 1.11 is now released with preliminary support for Go modules.
If a repository is checked outside of the GOPATH, then the Go toolchain will
use the module support by default. However, it requires a go.mod file so that
the toolchain can derive the import paths of all packages living under the
tree containing the go.mod file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,1 @@
+module github.com/google/go-cmp


neild:
Plan on tagging a 1.0 as well?

dsnet:
Probably not at this time. There are still some features and rough edges that I want to round off. I'll circle back to thinking about `cmp` more after the protobuf work.

Issue #40, opened 2017-09-10
Implement an AllowAllUnexported Option


jeromefroe:
Hi, I was wondering if you would consider adding a new option which would apply the `AllowUnexported` Option to all types and would not require a slice of input like the current `AllowUnexported` function does? 

My use case is that I have a lot of tests which use the `github.com/stretchr/testify` package, which ultimately uses `reflect.DeepEqual` for comparing two objects, that break in Go 1.9 because they try to compare structs which contain `time.Time` fields. Admittedly, I could define an `Equal` method for all such structs but that doesn't seem to address the root of the problem to me, which is that I would like `reflect.DeepEqual` functionality (which compares all fields, both private and public) with the caveat that I would like to check if the types being compared have an `Equal` method for testing equality. Furthermore, while for structs I could pass the structs being compared to `AllowUnexported`, I would like the same behavior for composite types as well (e.g. maps and slices) which is not supported by `AllowUnexported` currently since it specifically checks that the types of the objects passed to it are structs.

I would be more than happy to put up a PR for such an `Option`, a first look at the code leads me to believe that we could define an additional field on `state` indicating that we want to always check private fields. Then, when set to true, we would always extract a struct's fields in `tryExporting`. With that being said, I got the impression that you might be opposed to such an `Option` given the warnings for `AllowUnexported` already and so I thought it better to open an issue first where we could discuss.

Thanks!

dsnet:
`AllowUnexported` was a controversial option during the development of `cmp`. When dealing with unexported fields, there are several possible policies:

1. Ignore unexported by default; require use of `AllowUnexported` to actually compare them
2. Panic on unexported by default; require use of `AllowUnexported` to actually compare them
3. Always check unexported by default (`AllowUnexported` doesn't exist)

Policy 1 would cause failing tests to unexpectedly pass because `cmp.Equal` is not comparing a field the user thought they were comparing. So this idea was dropped early on.

Policy 2 (the camp I fall under), goes by the philosophy that the test writer should know explicitly what they are comparing, and provide an explicit whitelist (via `AllowUnexported`). Note that an unexported field added to a struct that formerly had only exported fields can cause `cmp` to panic. However, analysis of code history showed that this was rare, and if it did happen, I believe the test writer should think carefully about how they want to handle this failure. Fixing the test for these rare cases is probably a 1-line change to where they invoke `cmp.Equal`, I am okay with (before `cmp`, I often had to write dozens if not hundreds of lines to fix a bad comparison).

[...]

PR #97, opened 2018-11-12
add a DeepAllowUnexported cmpopt

mikedanese:
With a big note on its usage. This can sometimes be useful. We would like to migrate some of the diffing functionality we reimplement in k8s.io/kubernetes to use this library:

https://github.com/kubernetes/kubernetes/blob/3bcbc5da79978fd48525af82da03434f81e1cdd9/staging/src/k8s.io/apimachinery/pkg/util/diff/diff.go#L172

This option would give us a backwards compatible of our current API and allow us to move off of it incrementally and onto direct usage of cmp. We are unsure whether to copy this utility to k8s or add it here.

Fixes: https://github.com/google/go-cmp/issues/40

cc @awly

### DIFF
--- a/cmp/cmpopts/deep_allow_unexported.go
+++ b/cmp/cmpopts/deep_allow_unexported.go
@@ -1,0 +1,65 @@
+// Copyright 2018, The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE.md file.
+
+package cmpopts
+
+import (
+	"reflect"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+// DeepAllowUnexporeted creates a cmp.Option that recursively allows unexported
+// fields in the passed types.
+//
+// NOTE: Prefer to explicitly allow structs with internal types via
+// `cmp.AllowUnexported` rather than using this option. The philosophy behind this
+// is that the test writer should know explicitly what they are comparing, and
+// provide an explicit whitelist
+func DeepAllowUnexported(vs ...interface{}) cmp.Option {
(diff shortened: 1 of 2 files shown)


dsnet:
I believe this should be part of k8s.

> Is it a frequently requested feature?

In this situation, the number of thumbs up is a biased metric. Use of `DeepAllowUnexported` favors test writers who want to write tests fast without thinking carefully about whether they should compare upon unexported fields of some type that they don't own. It's use is detrimental to library authors since it means that users are now unintentionally depending on internal implementation details of a library that the author did not intend for users to access. Since there are almost always more users of a library than the authors of a library, the numbers are naturally skewed.

> Does the benefit outweigh the downside?

That's debatable.
* Not having `DeepAllowUnexported` forces people to think carefully about which types they should have unexported access to.
* When people use `DeepAllowUnexported` pervasively without thought, it leads to a situation that I saw not long ago, where adding an unexported field to one my types, and it broke thousands of tests that were using `reflect.DeepEqual` on that type.

In my opinion, the burden of k8s carrying their own `DeepAllowUnexported` is less than the dangers of having it as a feature in this repo.

dsnet:
Thanks for the PR, but I'm going to close this.

If we do provide official support for this in this repo, it would probably look something more like https://github.com/google/go-cmp/issues/40#issuecomment-438820128. For your use case in https://github.com/kubernetes/kubernetes/pull/70929, you could imagine doing `AllowUnexported(PackagePath("k8s.io"))` in the future to handle all types under a repo.

mikedanese:
Sounds good. Thanks!

PR #107, opened 2019-02-17
 Reorder logic to be consistent

dsnet:
Reorder logic to consistently handle steps in the order of:
	StructField, SliceIndex, MapIndex, Indirect, TypeAssertion

### DIFF
--- a/cmp/path.go
+++ b/cmp/path.go
@@ -33,6 +33,13 @@ type (
 		isPathStep()
 	}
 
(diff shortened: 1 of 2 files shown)


cmp/path.go
```diff
@@ -188,17 +188,6 @@ type (
 		typ reflect.Type
 	}
```
cybrcodr: Curious, why not have the methods right below each of these type definitions as well?

PR #108, opened 2019-02-17
Always ignore _ fields

dsnet:
The _ field in Go is special in that it cannot be read from or written to.
Thus, skip it by default. It is pointless to require the user to use
cmpopts.IgnoreUnexported just for the _ field.


dsnet:
It's useful when dealing with `unsafe` and you want to explicitly control the memory layout of the struct.

For example:
```go
type Foo struct {
     _ uint64
     v uint64 // lives at offset +8B
}
```

cybrcodr:
Neat.

I didn't know that _ field will take up space of given type (unless 0).

Good to know of another way to make a struct incomparable apart from embedding a func type.

PR #111, opened 2019-02-17
Add the Travis script to explicitly test intervening Go versions

dsnet:


.travis.yml
```diff
@@ -3,13 +3,21 @@ language: go
...
+        - go test -v -race ./...
+    - go: 1.9.x
+      script:
+        - go test -v -race ./...
+    - go: 1.10.x
+      script:
+        - go test -v -race ./...
+    - go: 1.11.x
       script:
         - diff -u <(echo -n) <(gofmt -d .)
```
cybrcodr: Just curious.  What does this do?  `gofmt -d` displays the diff.  comparing the diff output to ??

dsnet: I copied this line a long time ago from elsewhere. I believe the principle is that if the developer forgot to run `gofmt`, this would produce a non-empty output, thus failing the Travis script.

cybrcodr: Hah.  Makes sense now.  It seems like `gofmt -d` will always have exit 0 and hence to check if there is a delta or not, compare it's output against echo -n which does not produce anything.  Neat.

dmitshur: > I copied this line a long time ago from elsewhere.

#12. :)

PR #112, opened 2019-02-17
Refactor reporter implementation

dsnet:
Switch the internal interface for reporting to use a push-pop mechanism.
This API is strictly more flexible than the previous Report-only API since it
enables the reporter to re-construct the entire value tree.

This change is done in preparation for a major overhaul of reporter logic.


dsnet:
Merging... Got verbal LGTM from @cybrcodr

PR #115, opened 2019-02-20
Nudge people to use custom comparers

LMMilewski:
Currently we put push users towards options that handle unexported fields. A common scenario is:
  1. someone uses cmp.Diff on big.Int (transitively)
  2. they get a message that advertises Ignore/Allow Unexported options
  3. they see that AllowUnexported is hard to use correctly
  4. they use IgnoreUnexported
They end up with:
  cmp.IgnoreUnexported(big.Int{})
Which definitely doesn't do what's intended.

If we point out that a custom comparer is what they most likely need, then they are more likely to use cmp correctly.

### DIFF
--- a/cmp/cmpopts/ignore.go
+++ b/cmp/cmpopts/ignore.go
@@ -112,6 +112,10 @@ func (tf ifaceFilter) filter(p cmp.Path) bool {
 // In particular, unexported fields within the struct's exported fields
 // of struct types, including anonymous fields, will not be ignored unless the
 // type of the field itself is also passed to IgnoreUnexported.
+//
+// Avoid ignoring unexported fields of a type which you do not control (i.e. a
+// type from another repository), as changes to the implementation of such types
+// may change how the comparison behaves. Prefer a custom Comparer instead.
 func IgnoreUnexported(typs ...interface{}) cmp.Option {
 	ux := newUnexportedFilter(typs...)
 	return cmp.FilterPath(ux.filter, cmp.Ignore())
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1 +1,3 @@
 module github.com/google/go-cmp
+
+go 1.12
```
dsnet: Revert

cmp/options.go
```diff
@@ -202,7 +202,7 @@ type invalid struct{ core }
 
 func (invalid) filter(_ *state, _, _ reflect.Value, _ reflect.Type) applicableOption { return invalid{} }
 func (invalid) apply(s *state, _, _ reflect.Value) {
-	const help = "consider using AllowUnexported or cmpopts.IgnoreUnexported"
+	const help = "consider using Comparer. If you control the type, also consider AllowUnexported or cmpopts.IgnoreUnexported"
```
dsnet: `consider using a custom Comparer; if you control the implementation of type, you can also consider AllowUnexported or cmpopts.IgnoreUnexported`

cmp/cmpopts/ignore.go
```diff
@@ -112,6 +112,10 @@ func (tf ifaceFilter) filter(p cmp.Path) bool {
 // In particular, unexported fields within the struct's exported fields
 // of struct types, including anonymous fields, will not be ignored unless the
 // type of the field itself is also passed to IgnoreUnexported.
+//
+// Avoid ignoring unexported fields of a type which you do not control (i.e. a
+// type from another package), as changes to the implementation of such types
```
dsnet: s/package/repository/g

Repositories are generally the right boundary for ownership, not individual Go packages.

go.mod
LMMilewski: done

cmp/cmpopts/ignore.go
LMMilewski: done

cmp/options.go
LMMilewski: done

LMMilewski:
done, PTAL

PR #116, opened 2019-02-25
Support PackagePrefix in AllowUnexported

dsnet:
For sufficiently large codebases with many struct types defined,
it is untenable to specify a large list of all types in the repository
that AllowUnexported permits visibility for.

In Go, we observe that code with a shared owner often lives in the same
repository, and all sub-packages within a repository often share a common
package path prefix. Thus, permit expressing visibility allowances in terms
of package path prefixes instead of individual types.

While not explicitly documented, this feature allows the universal comparing
of unexported fields on all types:
	cmp.Equal(x, y, cmp.AllowUnexported(cmp.PackagePrefix("")))

This "feature" is intentionally undocumented (but a natural side-effect of the
expressed behavior of PackagePrefix) since it considered to be bad practice.


dsnet:
A problem with PackagePrefix is that it assumes that package paths are universally stable. This is pretty much true in open-source (otherwise it'd be impossible for packages to depend on each other), but may present a problem for companies that mangle the package path when vendoring 3rd party libraries.

cmp/options.go
```diff
@@ -338,17 +338,19 @@ func (cm comparer) String() string {
...
-// implementation of some external package may cause the result of Equal
-// to unexpectedly change. However, it may be valid to use this option on types
-// defined in an internal package where the semantic meaning of an unexported
-// field is in the control of the user.
+// Users must understand that comparing unexported fields from external packages
+// that are not owned by the user is unsafe since changes in the internal
+// implementation of external packages may cause the result of Equal to
+// unexpectedly change. However, it may be valid to use this option on types
+// owned by the user where the semantic meaning of an unexported field is in
+// full control of the user.
```
neild: Proposed simplification:

// Comparing unexported fields from packages that are not owned by the user
// is unsafe since changes in the internal implementation may cause the result
// of Equal to unexpectedly change. This option should only be used on types
// where the semantic meaning of unexported fields is in full control of the user.

cmp/options.go
```diff
@@ -363,24 +365,75 @@ func (cm comparer) String() string {
 //
 // In other cases, the cmpopts.IgnoreUnexported option can be used to ignore
 // all unexported fields on specified struct types.
-func AllowUnexported(types ...interface{}) Option {
+func AllowUnexported(templates ...interface{}) Option {
```
neild: Instead of overloading the meaning of the parameter, how about a new AllowUnexportedPackagePrefix option?

cmp/options.go
dsnet: Done.

cmp/options.go
dsnet: We're already lacking type safety here, so meh. I slightly prefer having it overloaded so that we can keep the documentation in one place.

neild: The documentation is in one place, but it's more confusing--I'm not concerned about type safety, but having this function do quite different things depending on what types you pass to it is muddled. "This function allows unexported fields on a type, that function allows unexported fields in a package (prefix)" is much clearer.

dsnet: Done. I named it "AllowUnexportedWithinPackages"

neild:
LGTM. (Is there no button to press for LGTM? If so, I can't find it. I miss Gerrit.)

dsnet:
Apologies for the flurry of changes; was struggling to find a good name. I settled on `AllowUnexportedWithinModule`. A module is defined as "a collection of related Go packages that are versioned together as a single unit", which seems to be exactly what we want here.

neild:
Hm. "Module" has a very specific definition, however, revolving around go.mod files. I think we should stay away from it unless we really mean it.

"PackagePrefix" is verbose, but seems pretty clear.

warpfork:
I will be fond of this feature when it lands.  :)

+1 towards "PackagePrefix" for naming.  I saw the word "Module" in the commit message and had a brief frission of shock and confusion (e.g. how and why would a library like this know about and read 'go.mod' files?) until reading the full message and patch.

PR #120, opened 2019-02-26
Update to go1.12

dsnet:


go.mod
```diff
@@ -1 +1,3 @@
 module github.com/google/go-cmp
+
+go 1.8
```
dmitshur: If you don't mind a question about this, can you share what criteria you used to select the Go version specified here?

PR #131, opened 2019-03-12
Implement specialized diffing for slices

dsnet:
Lists of primitives are a common-enough data structure that
it is worth providing specialized diffing for.

This provides significantly better readability for strings
and byte slices. There is also a heuristic for detecting
what a string should be diffed as a multiline string.

### DIFF
--- a/cmp/report_slices.go
+++ b/cmp/report_slices.go
@@ -1,0 +1,333 @@
+// Copyright 2019, The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE.md file.
+
+package cmp
+
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/google/go-cmp/cmp/internal/diff"
+)
+
+// CanFormatDiffSlice reports whether we support custom formatting for nodes
+// that are slices of primitive kinds or strings.
+func (opts formatOptions) CanFormatDiffSlice(v *valueNode) bool {
+	switch {
+	case opts.DiffMode != diffUnknown:
+		return false // Must be formatting in diff mode
+	case v.NumDiff == 0:
+		return false // No differences detected
+	case v.NumIgnored+v.NumCompared+v.NumTransformed > 0:
+		// TODO: Handle the case where someone uses bytes.Equal on a large slice.
+		return false // Some custom option was used to determined equality
+	case !v.ValueX.IsValid() || !v.ValueY.IsValid():
+		return false // Both values must be valid
+	}
+
+	switch t := v.Type; t.Kind() {
+	case reflect.String:
+	case reflect.Array, reflect.Slice:
+		// Only slices of primitive types have specialized handling.
+		switch t.Elem().Kind() {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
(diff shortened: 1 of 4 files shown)


cmp/report_slices.go
```diff
@@ -0,0 +1,333 @@
...
+	switch {
+	// If the text appears to be multi-lined text,
+	// then perform differencing across individual lines.
+	case isLinedText:
+		ssx := strings.Split(sx, "\n")
+		ssy := strings.Split(sy, "\n")
+		list = opts.formatDiffSlice(
+			reflect.ValueOf(ssx), reflect.ValueOf(ssy), 1, "line",
+			func(v reflect.Value, d diffMode) textRecord {
+				s := formatString(v.Index(0).String())
```
KrzysztofMadejski: For me the diffed text would be a lot readable if it wouldn't be quoted and packed in `strings.Join({`. Is it possible to specify an option to return text like that? 

I understand that the default is to produce valid go, hence the strings.Join and individual quoting. It would be interesting to know why did you take such design decision?

KrzysztofMadejski: And the other way around: would it be possible to force comparing string in one line even if it is long?

dsnet: > Is it possible to specify an option to return text like that?

For the time being, we're avoiding the addition of options that configure how the reporter displays diffs as there's an endless number of possible outputs. The moment you add an option, you are stuck supporting it forever. We prefer to use heuristics to choose a reasonable output. It's not perfect, but hopefully good enough most of the time. If you feel the current heuristics are sub-par, it's good to give examples where it's output isn't great.

> It would be interesting to know why did you take such design decision?

The reporter output prioritizes being unambiguous over being aesthetically nice. Not escaping the strings certainly looks nicer and may be easier to read in most situations, but there will be some set of strings where it is ambiguous whether the set of characters being displayed belong to the actual string itself or is part of the diff control characters (e.g., `-` and `+`).

It's possible to design an output scheme that preserve aesthetic niceness, while being ambiguous, but my fear is that this output will be unfamiliar to users. While the current output is not aesthetically nice, it is familiar to users.

If you have an idea of how to diff strings that is: 1) always unambiguous, 2) aesthetically pleasing, and 3) reasonably familiar to someone seeing it for the first time, we're happy to consider it. 

> And the other way around: would it be possible to force comparing string in one line even if it is long?

By default it does compare strings using the `==` operator regardless of the length or contents. Did you mean for the reporter? There's a difference between how `cmp` *compares* two values and how it *reports* the difference between two values. As mentioned earlier, we are intentionally avoiding options to configure how the reporter output looks.

However, there is the ability to add a [your own custom reporter](https://godoc.org/github.com/google/go-cmp/cmp#Reporter), so if you're not fond of the current output, you can always write your own.

KrzysztofMadejski: Great that there is an option to write custom reporter. Thanks for mentioning that.

It seems to me that using + and - in diff is unambigous. It is how `git diff` works, so I believe it is very familiar to devs. The first char of every line is either "+", "-" or a space. Then the actual contents follow. If actual contents contain "+" as a first char in diff it will be on the second:

In below example I prepend "+" before "import". 
```diff
~/Projects/handlertest$ git diff
diff --git a/body.go b/body.go
index 9b9ee88..1e51c6c 100644
--- a/body.go
+++ b/body.go
@@ -1,6 +1,6 @@
 package handlertest
 
-import (
++import (
        "bytes"
        "fmt"
        "io"
```

PR #142, opened 2019-05-26
Invoke String when formatting map keys 

### LINKED ISSUE #141: stringer not used for map keys
When rendering type `T`, the cmp package uses the String method for slice values, but not for map keys.

```
package main

import (
	"fmt"

)

type T [5]byte

func (t T) String() string { return string(t[:]) }

func main() {
	var t T
	copy(t[:], "world")
	m := map[T]string{t: "val"}

	s := []T{t}

	fmt.Println(cmp.Diff(nil, m))
	fmt.Println(cmp.Diff(nil, s))
}
```

Outputs:

```
  interface{}(
+ 	map[main.T]string{{0x77, 0x6f, 0x72, 0x6c, 0x64}: "val"},
  )

  interface{}(
+ 	[]main.T{s"world"},
  )
```

Try it here: https://play.golang.org/p/GS8f7ABvd61

If you agree this is a bug, I can take a crack at fixing it.

crawshaw:
This reverts a change introduced in commit 2940eda701 where the
cmp package stopped calling the String method when printing map
keys. The motivation for the change is unclear, indeed there was
a pre-existing test that String was called on map keys that the
commit changed.

Fixes #141

### DIFF
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -518,7 +518,7 @@ func comparerTests() []test {
 		y:     map[*pb.Stringer]*pb.Stringer(nil),
 		wantDiff: `
   map[*testprotos.Stringer]*testprotos.Stringer(
-- 	{⟪0xdeadf00f⟫: s"world"},
+- 	{s"hello": s"world"},
 + 	nil,
   )
 `,
--- a/cmp/report_reflect.go
+++ b/cmp/report_reflect.go
@@ -208,7 +208,6 @@ func (opts formatOptions) FormatValue(v reflect.Value, m visitedPointers) (out t
 func formatMapKey(v reflect.Value) string {
 	var opts formatOptions
 	opts.TypeMode = elideType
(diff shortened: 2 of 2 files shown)


dsnet:
I remember now why I avoided `String` for map keys. It was because of the following situation:
```go
type Int int
func NewInt(i int) *Int { return (*Int)(&i) }
func (i *Int) String() string { return strconv.Itoa(int(*i)) }

m := map[*MyInt]string{
    NewInt(5): "foo",
    NewInt(5): "bar",
}
```
The stringer for both keys prints to the same thing, which is slightly different from what the map is comparing on (the pointers).

However, that situation should be fixed by a more general change in the future that avoids stringer altogether if the outputs are ambiguous:
https://github.com/google/go-cmp/blob/6f77996f0c42f7b84e5a2b252227263f93432e9b/cmp/report_compare.go#L18-L20

PR #150, opened 2019-08-01
Refactoring by linters

osamingo:
The title said all.


osamingo:
I'll fix CI problem. and I'll retry re-PR after resolve it.

PR #168, opened 2019-10-27
Added capacity to slice creation, when capacity is known

AdityaHarindar:
Micro-optimization to reduce memory allocations, copying.


dsnet:
Thanks for the PR! However, it's our style to prefer natural Go first and only do pre-sized allocations if the code path is proven to be a performance bottleneck. None of these changes seem to be on the hot path, so the readability loss is not worth it.

PR #189, opened 2020-02-27
Document the test-only intentions of this package

dsnet:
This package was never intended to be used in production code.
Document the expectation that this package was only intended
to be used for writing tests.


cmp/compare.go
```diff
@@ -6,6 +6,9 @@
 //
 // This package is intended to be a more powerful and safer alternative to
 // reflect.DeepEqual for comparing whether two values are semantically equal.
+// It is intended to only be used in tests, as cmp may panic if it cannot
+// compare the values. It's propensity towards panicking means that it is
```
neild: Its

cmp/compare.go
```diff
@@ -6,6 +6,9 @@
...
+// unsuitable for production environments where a spurious panic may be fatal.
```
neild: Maybe also mention that performance is not really a goal?

cmp/compare.go
dsnet: Kept it in untracted form for the first occurrence. Used contracted form for later occurences.

cmp/compare.go

cmp/compare.go
```diff
@@ -6,6 +6,10 @@
 //
 // This package is intended to be a more powerful and safer alternative to
 // reflect.DeepEqual for comparing whether two values are semantically equal.
+// It is intended to only be used in tests, as performance is not a goal and
+// it may panic if it cannot compare the values. It's propensity towards
```
neild: s/It's/Its/

cmp/compare.go
dsnet: Oh, you were probably referring to the "it's", which is incorrect grammar. Sigh, the inability to highlight a specific word in GitHub PR is confusing.

cmp/compare.go
dsnet: Just realized this. Already fixed.

PR #200, opened 2020-05-13
Refactor tests to use golden test files

dsnet:
Refactor the unit tests to read the diffs from a file
rather than being stored as a Go string literal.
The advantage of using a golden file is to ease
updating the diffs whenever the reporter output is changed.


dsnet:
\cc cybrcodr for review. I don't know why I can't assign you as reviewer anymore.

dsnet:
> Ok, I got back now.

That was fast.

PR #201, opened 2020-05-16
Avoid custom diffing for nil slices

### LINKED ISSUE #157: Unexpected diffs of nil slices
For some values (types?), `cmp.Diff` returns no apparent difference though `cmp.Equal` returns false. The following program shows no `-/+` for `[]int`, `[]byte`, and `[]float64` types.

`Diff` does produce output, so I'm not certain this is a bug. I was expecting (and would grateful) to see `- nil / + {}` in these cases as well.

```go
package main

import (
	"fmt"
	"reflect"

)

func main() {
	for _, tt := range []struct{ Nil, Empty interface{} }{
		{Nil: []int(nil), Empty: []int{}},
		{Nil: []byte(nil), Empty: []byte{}},
		{Nil: []string(nil), Empty: []string{}},
		{Nil: []float64(nil), Empty: []float64{}},
		{Nil: []struct{}(nil), Empty: []struct{}{}},
		{Nil: []interface{}(nil), Empty: []interface{}{}},
	} {
		fmt.Println("Type:", reflect.TypeOf(tt.Nil).String())
		fmt.Println("DeepEqual:", reflect.DeepEqual(tt.Nil, tt.Empty))
		fmt.Println("cmp.Equal:", cmp.Equal(tt.Nil, tt.Empty))
		fmt.Println("Diff:\n", cmp.Diff(tt.Nil, tt.Empty))
		fmt.Println("=====")
	}
}
```
Output:
```
Type: []int
DeepEqual: false
cmp.Equal: false
Diff:
   []int{}

=====
Type: []uint8
DeepEqual: false
cmp.Equal: false
Diff:
   bytes.Join({}, "")

=====
Type: []string
DeepEqual: false
cmp.Equal: false
Diff:
   []string(
- 	nil,
+ 	{},
  )

=====
Type: []float64
DeepEqual: false
cmp.Equal: false
Diff:
   []float64{}

=====
Type: []struct {}
DeepEqual: false
cmp.Equal: false
Diff:
   []struct{}(
- 	nil,
+ 	{},
  )

=====
Type: []interface {}
DeepEqual: false
cmp.Equal: false
Diff:
   []interface{}(
- 	nil,
+ 	{},
  )

=====
```

at-ishikawa:
Proposal to fix #157 

I'm not sure what the best format for these cases are, so just stop using custom format for slices as a proposal.
See `cmp/testdata/diffs` for the example diff for nil cases.


dsnet:
Thanks! Falling back on the typical diffing behavior seems reasonable.

PR #205, opened 2020-05-31
Create a simple example for the IgnoreFields

colinnewell:
I was struggling to figure out how to use IgnoreFields initially so I figured an example would be helpful.  I've essentially ripped off the example from the parent module with a proper copy paste job for now with a quick example.

If you can suggest a better way to do it, I'd be happy to clean up the code.

I suspect I could also make the example clearer, but I'd quite like to clean up the code before going too far.

### DIFF
--- a/cmp/cmpopts/example_test.go
+++ b/cmp/cmpopts/example_test.go
@@ -7,7 +7,12 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/go-cmp/cmp/internal/flags"
 )
+
+func init() {
+	flags.Deterministic = true
+}
 
 // Use IgnoreFields to ignore fields on a type when comparing.
 // Provide an interface of the type, and the field names to ignore.
@@ -28,7 +33,7 @@ func ExampleIgnoreFields_testing() {
 	//   	SSID:      "CoffeeShopWiFi",
 	// - 	IPAddress: s"192.168.0.2",
 	// + 	IPAddress: s"192.168.0.1",
-	//   	NetMask:   net.IPMask{0xff, 0xff, 0x00, 0x00},
+	//   	NetMask:   {0xff, 0xff, 0x00, 0x00},
 	//   	Clients: []cmpopts_test.Client{
 	//   		... // 3 identical elements
 	//   		{Hostname: "espresso", ...},
@@ -40,6 +45,7 @@ func ExampleIgnoreFields_testing() {
 	// + 		},
 	//   	},
 	//   }
+
 }
 
 type (
--- a/cmp/cmpopts/example_test.go
+++ b/cmp/cmpopts/example_test.go
@@ -18,8 +18,9 @@ func init() {
 	flags.Deterministic = true
 }
 
-// Use IgnoreFields to ignore fields on a type when comparing.
-// Provide an interface of the type and the field names to ignore.
+// Use IgnoreFields to ignore fields on a struct type when comparing
+// by providing a value of the type and the field names to ignore.
+// Typically, a zero value of the type is used (e.g., foo.MyStruct{}).
 func ExampleIgnoreFields_testing() {
 	// Let got be the hypothetical value obtained from some logic under test
 	// and want be the expected golden data.
--- a/cmp/cmpopts/example_test.go
+++ b/cmp/cmpopts/example_test.go
@@ -1,0 +1,119 @@
+package cmpopts_test
+
+import (
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+)
+
+// Use IgnoreFields to ignore fields on a type when comparing.
+// Provide an interface of the type, and the field names to ignore.
+func ExampleIgnoreFields_testing() {
+	// Let got be the hypothetical value obtained from some logic under test
+	// and want be the expected golden data.
+	got, want := MakeGatewayInfo()
+
+	// Note that the Diff will still potentially show ignored fields as different,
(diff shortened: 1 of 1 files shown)


frioux:
This seems like a solid PR to me.  The only problem is that cmp explicitly thwarts tests like this by injecting randomness into the output format.  @colinnewell you might set the flag that disables that for your test and then it'd pass.

colinnewell:
Thanks for the comments @frioux.  Did you see where this randomness is coming from?  I'm failing to spot it so far.  One of the reasons I didn't push this PR too far was that it was pretty much a rip off of `cmp/example_test.go` which seemed ugly.  The other thing that I'm struggling to figure out is why that one doesn't suffer from the same problem mine does.

If figure I may as well try to get the tests to pass consistently rather than randomly.

colinnewell:
Ah, found it: `cmp/report_text.go`.

colinnewell:
Thanks for the prod.  I've rebased and fixed the test so that it runs consistently.

cmp/cmpopts/example_test.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Use IgnoreFields to ignore fields on a type when comparing.
+// Provide an interface of the type, and the field names to ignore.
+func ExampleIgnoreFields_testing() {
+	// Let got be the hypothetical value obtained from some logic under test
+	// and want be the expected golden data.
+	got, want := MakeGatewayInfo()
+
+	// Note that the Diff will still potentially show ignored fields as different,
```
dsnet, line 24: > will still potentially show ignored fields as different,

Is still true? I recall fixing the reporter to stop printing ignored fields and using `...` instead, even for adjacent fields.

cmp/cmpopts/example_test.go
```diff
@@ -0,0 +1,125 @@
+package cmpopts_test
```
dsnet, line 1: Please add license header.

cmp/cmpopts/example_test.go
```diff
@@ -0,0 +1,125 @@
...
+	//   		{Hostname: "espresso", ...},
+	//   		{Hostname: "latte", LastSeen: s"2009-11-10 23:00:23 +0000 UTC", ...},
+	// + 		{
+	// + 			Hostname:  "americano",
+	// + 			IPAddress: s"192.168.0.188",
+	// + 			LastSeen:  s"2009-11-10 23:03:05 +0000 UTC",
+	// + 		},
+	//   	},
+	//   }
+
```
dsnet, line 48: Drop blank line.

cmp/cmpopts/example_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/go-cmp/cmp/internal/flags"
+)
+
+func init() {
+	flags.Deterministic = true
+}
+
+// Use IgnoreFields to ignore fields on a type when comparing.
+// Provide an interface of the type, and the field names to ignore.
```
dsnet, line 18: Nit: drop the comma before the "and" as "the field names to ignore" is not a complete sentence.

dsnet (commented):
Thanks!

cmp/cmpopts/example_test.go
dsnet: https://github.com/google/go-cmp/blob/d713870ac17fdb9ee5e2ee48ff6562dfb1c0157b/cmp/report_compare.go#L258-L265

colinnewell: The example itself demonstrates what I was trying to say.  I ignore `IPAddress` but it still shows up in the diff.  I assume that the reason I added 'potentially' to the sentence is because I had come across your behaviour to minimise that.

If you do end up changing that behaviour, the example will highlight it so it should be easy to change that comment then.

Alternatively, is it the word 'adjacent' that you're objecting to?

cmp/cmpopts/example_test.go
colinnewell: Done

cmp/cmpopts/example_test.go
colinnewell: Done

cmp/cmpopts/example_test.go
colinnewell: Done

cmp/cmpopts/example_test.go
dsnet: I know people aren't fond of 2020, but that's the current year ;)

cmp/cmpopts/example_test.go
dsnet: I think there's a crucial difference between "in the diff" as you just said versus "as different" as currently written in the example. The former denotes an issues with display representation, while the latter subtly suggests something about the semantics of how the comparison is performed.

Perhaps this is better worded as:
```go
// While the specified fields will be semantically ignored for the comparison,
// the fields may be printed in the diff when displaying entire values
// that are already determined to be different.
```

cmp/cmpopts/example_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/google/go-cmp/cmp/internal/flags"
+)
+
+func init() {
+	flags.Deterministic = true
+}
+
+// Use IgnoreFields to ignore fields on a type when comparing.
+// Provide an interface of the type and the field names to ignore.
```
dsnet, line 22:
```go
// Use IgnoreFields to ignore fields on a struct type when comparing
// by providing a value of the type and the field names to ignore.
// Typically, a zero value of the type is used (e.g., foo.MyStruct{}).
```

colinnewell: Used, thanks.

PR #215, opened 2020-06-11
Limit verbosity of reporter output

dsnet:
A common complaint is that the reporter it prints out too much
irrelevant information, resulting in a low signal-to-noise ratio.
Improve this metric by imposing a verbosity limit. For nodes that
are equal, we set the verbosity level is a lower value than when
the nodes are inequal.

Other minor changes:
* Adjust heuristic for triple-quote usage to operate on more cases.
* Elide type more aggressively for equal nodes.
* Printing the address for a slice includes the length and capacity.
* The pointed-at value for map keys are printed.

### DIFF
--- a/cmp/report_compare.go
+++ b/cmp/report_compare.go
@@ -11,10 +11,6 @@ import (
 	"github.com/google/go-cmp/cmp/internal/value"
 )
 
-// TODO: Enforce limits?
-//	* Enforce maximum number of records to print per node?
-//	* Enforce maximum size in bytes allowed?
-//	* As a heuristic, use less verbosity for equal nodes than unequal nodes.
 // TODO: Enforce unique outputs?
 //	* Avoid Stringer methods if it results in same output?
 //	* Print pointer address if outputs still equal?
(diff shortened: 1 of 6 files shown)


dsnet:
The verbosity levels could use more fine tuning, but this is good enough for now, and much better than before in my opinion.

Upcoming changes:
* Add logic for the reporter work harder on the output if it is ambiguous
* Add better cycle formatting logic. Right now, if a cycle is detected, the user has no idea where it occurred.

cmp/testdata/diffs
```diff
@@ -155,8 +155,8 @@
 >>> TestDiff/Comparer#42
 <<< TestDiff/Comparer#43
   map[*int]string{
-- 	⟪0xdeadf00f⟫: "hello",
-+ 	⟪0xdeadf00f⟫: "world",
+- 	&⟪0xdeadf00f⟫0: "hello",
```
cybrcodr, line 158: What/where is the ending 0 coming from in &⟪0xdeadf00f⟫0?

dsnet: It's the value of the pointed-at value. Previously, it was elided, which is confusing.

cmp/report_compare.go
```diff
@@ -71,10 +67,27 @@ func (opts formatOptions) WithTypeMode(t typeMode) formatOptions {
...
+func (opts formatOptions) WithVerbosity(level int) formatOptions {
+	opts.VerbosityLevel = level
+	opts.LimitVerbosity = true
+	return opts
+}
+func (opts formatOptions) verbosity() uint {
+	if opts.VerbosityLevel < 0 {
+		return 0
+	}
+	return uint(opts.VerbosityLevel)
```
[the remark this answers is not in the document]
dsnet: I set it to a maximum of 16, which is half the width of an `int` on a 32-bit system.

PR #217, opened 2020-06-13
Improve reporting of values with cycles

dsnet:
Previously, the reporter could handle formatting values with cycles
in that it did not crash with a stack overflow. However, the output
was not particularly understandable as it did not surface to the user
why a particular value was truncated, and if it was truncated due
to a cyclic reference, what was the referent.

This change annotates the reporter tree with pointer information
so that a later pass can inject reference information if it is needed
to produce more understandable output.

### DIFF
--- a/cmp/report_reflect.go
+++ b/cmp/report_reflect.go
@@ -12,7 +12,6 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	"github.com/google/go-cmp/cmp/internal/flags"
 	"github.com/google/go-cmp/cmp/internal/value"
 )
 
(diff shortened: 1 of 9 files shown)


dsnet:
This is the last major cmp change I have in mind. It's unfortunately a bit complex as cycle detection is hard. Printing cycles in a sensible way for humans is even harder.

cmp/report_reflect.go
```diff
@@ -75,26 +69,54 @@ func (opts formatOptions) FormatType(t reflect.Type, s textNode) textNode {
 			typeName = "(" + typeName + ")"
 		}
 	}
+	return &textWrap{Prefix: typeName, Value: wrapParens(s)}
+}
+
+// wrapParens wraps s with a set of parenthesis, but avoids it if it is super
```
cybrcodr, line 75: super?  Missing end phrase?

dsnet: Ooops. Fixed.

dsnet:
Thank you for reviewing all my changes. I think this concludes my series of planned cmp improvements.

PR #218, opened 2020-06-14
Update test case names

dsnet:
Provide a unique name for every test case.
Provide a reason for every test case.
The purpose of a unique name is so that insertion/removal of a
case does not cause all subsequent names to suddenly shift,
causing a larger number of differences in testdata/diffs.

### DIFF
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -155,9 +155,12 @@ func TestDiff(t *testing.T) {
 				gotDiff = cmp.Diff(tt.x, tt.y, tt.opts...)
 			}()
 
-			// TODO: Require every test case to provide a reason.
-			// TODO: Forbid any test cases with the same name.
-			if tt.wantPanic == "" {
+			switch {
+			case strings.Contains(t.Name(), "#"):
+				panic("unique test name must be provided")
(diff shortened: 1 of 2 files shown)


cmp/compare_test.go
```diff
@@ -1289,448 +1342,504 @@ func embeddedTests() []test {
...
 		x:     createStructG(0),
 		y:     createStructG(1),
 		opts: []cmp.Option{
 			cmp.AllowUnexported(ts.ParentStructG{}, privateStruct),
 		},
 		wantEqual: false,
+		reason:    "the two values differ on some fields",
 	}, {
-		label:     label + "ParentStructH",
+		label:     label + "ParentStructH/EqualNil",
```
cybrcodr, line 1674: Does the string also require / in front?  Same for the next 5 cases below?

I didn't see this in the diff file, so I wasn't sure.

dsnet: Nice catch, yea.

You can see it missing a "/" here:
```
>> TestDiff/EmbeddedStruct/ParentStructG/Inequal
<<< TestDiff/EmbeddedStructParentStructH/Inequal
```

PR #225, opened 2020-06-26
Drop xerrors since functionality is merged to errors in std lib

arvenil:
Drop the only dependency since all functionality from xerrors used in go-cmp is now merged to errors in standarad library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,6 @@ sudo: false
 language: go
 matrix:
   include:
-    - go: 1.8.x
-      script:
-        - go test -v -race ./...
-    - go: 1.9.x
-      script:
-        - go test -v -race ./...
-    - go: 1.10.x
-      script:
-        - go test -v -race ./...
-    - go: 1.11.x
-      script:
-        - go test -v -race ./...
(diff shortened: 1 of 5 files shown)


arvenil:
@googlebot I signed it!

dsnet:
Thanks for the PR, but we can't accept this. The new errors API was added in go1.14, but our current minimally supported version is go1.8. The `cmp` package is very widely used, so we need to support a wide range of Go versions. It will be a long time before we get to go1.14 has the minimally supported version.

arvenil:
Hi, makes sense, feel free to close it. However I'm suprised since people using older golang version can always download older go-cmp version 🤷.

dsnet:
> people using older golang version can always download older go-cmp version 🤷.

It's not quite that simple. Suppose there is module A at version v1.5 and that module depends on cmp at version v0.6 (with go1.14 as the minimum support).

If you're constrained to an old version of Go and want to use A@v1.5, you're stuck since it requires you to use cmp@v0.6, which requires that you use go1.14.

Closing this PR since we won't get to this point for at least a year or more.

PR #228, opened 2020-07-14
Fixed panic message in IgnoreUnexported

ko30005:
Fixed panic message more specifically.
With this fix, you can see that IgnoreUnexported only supports a struct without the user having to look at the code.


dsnet:
Thanks for the PR.

For consistency, we probably want to adjust this panic message too:

https://github.com/google/go-cmp/blob/d669b046d12237b504e86b93d6b25ec551e8c349/cmp/cmpopts/struct_filter.go#L43-L46

ko30005:
Thanks!
I fixed this panic message.

dsnet:
Thanks!

PR #229, opened 2020-07-20
Use triple-quote formatting for multiline strings

dsnet:
For strings, []bytes containing text data, Error method output, and
String method output, use the triple-quoted syntax.
This improves readability by presenting the data more naturally
compared to a single-line quoted string with many escaped characters.


dsnet:
\cc @sh0rez

PR #231, opened 2020-07-29
Update compare.go documentation

ernest-galbrun:
Fix the doc about the cmp.Diff method. It was mentioning the plus or minus sign being printed if, respectively, the field was added to x or removed to y, but both things are equivalent. The sign change if a field is removed or added from y alone.


dsnet:
Thanks!

PR #240, opened 2020-10-20
Adjust for reflect.Type.NumMethod change in Go1.16

dsnet:
In Go1.16, the reflect.Type.NumMethod method will no longer report
unexported fields, matching the documented behavior on the method.
This means that t.NumMethod() == 0 is no longer a reliable means
to detect whether an interface type is the empty interface or not.
Fix the code to check whether the empty interface itself implements
the target type.


dsnet:
See https://github.com/golang/go/issues/22075 for details.

neild (approved):

PR #243, opened 2020-11-13
Remove xerrors dependency and use stdlib.

googollee:


dsnet:
I apologize, but we can't accept this change. See https://github.com/google/go-cmp/pull/225#issuecomment-650431592

googollee:
I understand that it needs to support Go 1.8. As go mod has version control, could we raise the version number to drop the dependency? If users want to use it under Go 1.14, they could pin the version 0.5.x.

dsnet:
See https://github.com/google/go-cmp/pull/225#issuecomment-650434186

googollee:
Thanks for clarifying.

PR #251, opened 2021-02-04
cmp/cmpopts: use errors.Is with ≥go1.13 in compareErrors

tklauser:
Use the standard definition of errors.Is to implement compareErrors with
≥go1.13. Retain the implementation using golang.org/x/xerrors for versions <go1.13.

This will allow packages using newer Go versions and already relying on
the errors package to get rid of the transitive dependency on
golang.org/x/xerrors.

### DIFF
--- a/cmp/cmpopts/errors_go113.go
+++ b/cmp/cmpopts/errors_go113.go
@@ -1,0 +1,15 @@
+// Copyright 2021, The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build go1.13
+
+package cmpopts
+
+import "errors"
+
+func compareErrors(x, y interface{}) bool {
+	xe := x.(error)
+	ye := y.(error)
+	return errors.Is(xe, ye) || errors.Is(ye, xe)
+}
--- a/cmp/cmpopts/errors_xerrors.go
+++ b/cmp/cmpopts/errors_xerrors.go
@@ -1,0 +1,18 @@
+// Copyright 2021, The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !go1.13
+
+// TODO(≥go1.13): For support on <go1.13, we use the xerrors package.
+// Drop this file when we no longer support older Go versions.
+
+package cmpopts
+
+import "golang.org/x/xerrors"
+
+func compareErrors(x, y interface{}) bool {
+	xe := x.(error)
+	ye := y.(error)
+	return xerrors.Is(xe, ye) || xerrors.Is(ye, xe)
+}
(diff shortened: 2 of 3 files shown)


tklauser:
I see that previously #225 and #243 attempted to get rid of the golang.org/x/xerrors as well. However it looks like these weren't accepted based on the fact that they broke `cmp` for Go versions older than 1.13, while this PR retains compatibility for older versions.

cmp/cmpopts/errors_go113.go
```diff
@@ -0,0 +1,16 @@
...
+// +build go1.13
+
+package cmpopts
+
+import "errors"
+
+func compareErrors(x, y interface{}) bool {
+	xe := x.(error)
+	ye := y.(error)
+	// Use standard definition of errors.Is with ≥go1.13.
```
dsnet, line 14: I don't think this comment is necessary anymore.

cmp/cmpopts/errors_xerrors.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2021, The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !go1.13
```
dsnet, line 5: Add a TODO in this file to drop it when the minimally supported version of `cmp` is greater than or equal to Go 1.13.

cmp/cmpopts/errors_xerrors.go
```diff
@@ -0,0 +1,16 @@
...
+
+package cmpopts
+
+import "golang.org/x/xerrors"
+
+func compareErrors(x, y interface{}) bool {
+	xe := x.(error)
+	ye := y.(error)
+	// Use xerrors with <go1.13.
```
dsnet, line 14: I don't think this comment is necessary anymore.

cmp/cmpopts/errors_xerrors.go
dsnet: Something like:

// TODO(≥go1.13): For support on <go1.13, we use the xerrors package.
// Drop this file when we no longer support older Go versions.

dsnet:
What's the effect of this PR on the `go.mod` file?

tklauser:
> What's the effect of this PR on the `go.mod` file?

As far as I can tell there will be no effect on the `go.mod` file because `cmp/cmpopts/util_test.go` still uses `xerrors` unconditionally. Running `go mod tidy` doesn't change `go.mod` or `go.sum`.

However, a package importing `cmp` which uses `go1 .13` or newer in `go.mod` should no longer see an indirect dependency on `xerrors`.

dsnet:
The presence of `xerrors` in the `go.mod` file is unfortunate especially since older versions of Go (which is why the dependency even needs to exist) don't use Go modules.

Anyways, I agree this change in a step in the direction we want to go.

dsnet (approved):

Issue #244, opened 2020-11-20
Improve optimality of diffing algorithm


Eun:
Given following example:
```golang
func TestDiff(t *testing.T) {
	fmt.Println(cmp.Diff(
		// Hello Universe
		[]uint8{0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x20, 0x55, 0x6e, 0x69, 0x76, 0x65, 0x72, 0x73, 0x65},
		// Hello World
		[]uint8{0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x20, 0x57, 0x6f, 0x72, 0x6c, 0x64},
	))
}
``` 
sometimes reports
```
  []uint8{
  	... // 4 identical elements
  	0x6f,
  	0x20,
- 	0x55,
+ 	0x57,
- 	0x6e,
+ 	0x6f,
- 	0x69,
- 	0x76,
- 	0x65,
  	0x72,
- 	0x73,
+ 	0x6c,
- 	0x65,
+ 	0x64,
  }
```
sometimes
```
  bytes.Join({
  	"Hello ",
- 	"Universe",
+ 	"World",
  }, "")
```

_Originally posted by @Eun in https://github.com/google/go-cmp/issues/238#issuecomment-731204953_

dsnet:
Implementation note: the current diffing algorithm is guaranteed to run in O(n) based on a  greedy search algorithm. There are other diffing algorithms that produce diffs that are guaranteed to be optimal (i.e., have the [shortest Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance)), but run in O(n^2). We should consider improving the diffing algorithm to have some degree of back-tracking so that it doesn't get stuck in a sub-optimal solution. At the same time, it's an important goal that the diffing algorithm avoid O(n^2) behavior.

dsnet:
This appears to have been fixed by https://github.com/google/go-cmp/commit/449e17c6c9daf9b0c84a35fef7d79321b9535763

PR #258, opened 2021-04-30
Print as text if mostly text

### LINKED ISSUE #257: better diffs when most chars are ascii

dsnet:
The previous heuristic of treating strings as binary data
if it contains any invalid UTF-8 was too strict.
Loosen the heuristic to check if most of the characters
are printable text.

Fixes #257

### DIFF
--- a/cmp/report_slices.go
+++ b/cmp/report_slices.go
@@ -7,6 +7,7 @@ package cmp
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"reflect"
 	"strconv"
 	"strings"
@@ -96,16 +97,16 @@ func (opts formatOptions) FormatDiffSlice(v *valueNode) textNode {
 	}
 
 	// Auto-detect the type of the data.
-	var isLinedText, isText, isBinary bool
 	var sx, sy string
 	var ssx, ssy []string
+	var isString, isMostlyText, isPureLinedText, isBinary bool
 	switch {
 	case t.Kind() == reflect.String:
 		sx, sy = vx.String(), vy.String()
-		isText = true // Initial estimate, verify later
+		isString = true
 	case t.Kind() == reflect.Slice && t.Elem() == reflect.TypeOf(byte(0)):
 		sx, sy = string(vx.Bytes()), string(vy.Bytes())
-		isBinary = true // Initial estimate, verify later
+		isString = true
 	case t.Kind() == reflect.Array:
 		// Arrays need to be addressable for slice operations to work.
 		vx2, vy2 := reflect.New(t).Elem(), reflect.New(t).Elem()
@@ -113,13 +114,12 @@ func (opts formatOptions) FormatDiffSlice(v *valueNode) textNode {
 		vy2.Set(vy)
 		vx, vy = vx2, vy2
 	}
-	if isText || isBinary {
-		var numLines, lastLineIdx, maxLineLen int
-		isBinary = !utf8.ValidString(sx) || !utf8.ValidString(sy)
+	if isString {
+		var numTotalRunes, numValidRunes, numLines, lastLineIdx, maxLineLen int
 		for i, r := range sx + sy {
-			if !(unicode.IsPrint(r) || unicode.IsSpace(r)) || r == utf8.RuneError {
-				isBinary = true
-				break
+			numTotalRunes++
+			if (unicode.IsPrint(r) || unicode.IsSpace(r)) && r != utf8.RuneError {
+				numValidRunes++
 			}
 			if r == '\n' {
 				if maxLineLen < i-lastLineIdx {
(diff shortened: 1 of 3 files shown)


dsnet:
\cc @rogpeppe

cmp/testdata/diffs
```diff
@@ -1046,6 +1046,26 @@
 + 	},
   }
 >>> TestDiff/Reporter/LargeStandaloneString
+<<< TestDiff/Reporter/MostlyTextString
```
dsnet, line 1049: This output is suboptimal due to #244.

Ideally it would look like:
```
  strings.Join({
  	"org-4747474747474747,bucket-4242424242424242:m,tag1=a,tag2=aa",
- 	",\xff=_value",
  	" _value=2 11\norg-4747474747474747,bucket-4242424242424242:m,tag1=a,tag2=bb",
- 	",\xff=_value",
  	" _value=2 21\norg-4747474747474747,bucket-4242424242424242:m,tag1=b,tag2=cc",
- 	",\xff=_value",
  	" _value=1 21\norg-4747474747474747,bucket-4242424242424242:m,tag1",
  	"=a,tag2",
- 	",\xff=_value",
  	"=dd _value=3 31\norg-4747474747474747,bucket-4242424242424242:m,tag1",
- 	",\xff=_value",
  	"=c _value=4 41\n",
  }, "")
```

rogpeppe: all the same, this is _way_ better, thanks!

cmp/report_slices.go
```diff
@@ -128,8 +127,10 @@ func (opts formatOptions) FormatDiffSlice(v *valueNode) textNode {
 				numLines++
 			}
 		}
-		isText = !isBinary
-		isLinedText = isText && numLines >= 4 && maxLineLen <= 1024
+		isPureText := numValidRunes == numTotalRunes
+		isMostlyText = float64(numValidRunes)/float64(numTotalRunes) > 0.95
```
rogpeppe, line 131: ISTM that this only triggers on longer strings. For example, if there's one non-printable character in a 19 character string, it won't be treated as "mostly text". Perhaps there could be some kind of count cut-off too?

rogpeppe (approved):
Thanks very much for doing this so promptly. I have one thought for a possible improvement but otherwise LGTM.

cmp/report_slices.go
dsnet: The challenged with the heuristics is that they're never right 😞.
The benefit with heuristics is that they don't need to be right and we can do whatever is best for humans 😏.

> For example, if there's one non-printable character in a 19 character string, it won't be treated as "mostly text".

Did you mean something like?
```go
isMostlyText = float64(numValidRunes)/float64(numTotalRunes) > 0.95 || numTotalRunes-numValidRunes < 4
```

rogpeppe: yeah, something like that.

how about this?

```
	const mostlyTextProportion = 0.90

	isMostlyText := numTotalRunes == numValidRunes
	if !isMostlyText && numTotalRunes > 3 {
		threshold := int(math.Floor(mostlyTextProportion * float64(numTotalRunes)))
		isMostlyText = numValidRunes >= threshold
	}
```

That gives the following set of cutoff points (number of invalid characters / total number of characters):

```
0/1
0/2
0/3
1/4
1/5
1/6
1/7
1/8
1/9
1/10
2/11
2/12
2/13
2/14
2/15
2/16
2/17
2/18
2/19
2/20
3/21
3/22
3/23
3/24
3/25
3/26
3/27
3/28
3/29
3/30
4/31
4/32
4/33
4/34
4/35
4/36
4/37
4/38
4/39
4/40
5/41
5/42
5/43
5/44
5/45
5/46
5/47
5/48
5/49
5/50
6/51
6/52
6/53
6/54
6/55
6/56
6/57
6/58
6/59
6/60
7/61
7/62
7/63
7/64
7/65
7/66
7/67
7/68
7/69
7/70
8/71
8/72
8/73
8/74
8/75
8/76
8/77
8/78
8/79
8/80
9/81
9/82
9/83
9/84
9/85
9/86
9/87
9/88
9/89
9/90
10/91
10/92
10/93
10/94
10/95
10/96
10/97
10/98
10/99
```

dsnet: The heuristic seems fairly complicated. Since most of those ratios are below 80%, I think it would be easier and simpler to just reduce the ratio to 80%.

dsnet: https://play.golang.org/p/M6s-3G_Wg-T

rogpeppe: FWIW the heuristic was "round down from 90%", but I'm happy to go with 80% (which gives these numbers, I think: https://play.golang.org/p/lCIupl-0WMO).

neild (approved):

cmp/report_slices.go
dsnet: I went with math.Floor, but keep the threshold at 90%. We can adjust this in the future as necessary.

PR #266, opened 2021-07-19
Fix textual printing of byte slices

dsnet:
There are two bugs being fixed:

1.  The hueristic for whether a slice of byte looks like text
    should check whether a rune IsPrint OR IsSpace, and not both.
    Only a single rune (i.e., U+0020) ever satisfies both conditions.

    Previously, it would print as:
    ```
    MyBytes{0x68, 0x65, 0x6c, 0x6c, 0x6f}
    ```
    and now it would now print as:
    ```
    MyBytes(MyBytes("hello"))
    ```

2.  If we're printing as string, then we should set skipType=true
    since we already explicitly format the value with the type.

    Previously, it would print as:
    ```
    MyBytes(MyBytes("hello"))
    ```
    and now it would now print as:
    ```
    MyBytes("hello")
    ```


dsnet:
\cc @neild.

neild (approved):

PR #268, opened 2021-07-22
Use sha256 in test

dsnet:
Some aggressive dependency checks flag the use of md5.
Switch to sha256 as it accomplishes the same purpose.


dsnet:
\cc @neild FYI, I've been using "squash+merge" rather than just "merge" to keep the code history cleaner.

neild (approved):

PR #275, opened 2021-09-16
Reduce minimum length for specialize string diffing

dsnet:
The original threshold of 64 was chosen without much thought.
Lower it to 32 now that we have some concrete examples
that it is aesthetically better.


dsnet:
\cc @neild

neild (approved):

PR #282, opened 2021-12-07
Drop hacks to work around Go reflection bugs in Go1.9

dsnet:
Now that Go 1.11 is the minimally supported version,
we can drop some local hacks to work around bugs in
reflect that were present in Go1.9.


dsnet:
I've been waiting to submit this for quite a while :P

neild (approved):

PR #290, opened 2022-03-22
Add Go 1.18 support

Aoang:


.github/workflows/test.yml
```diff
@@ -19,12 +19,8 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Checkout code
       uses: actions/checkout@v2
-      with:
```
dsnet, line 22: This needs to be kept for compatibility with lower versions of Go

dsnet:
Thanks for the PR!

dsnet (changes requested):

dsnet:
\cc @neild. We're still supporting versions down to Go1.11? What's a reasonable lower bound to support these days? I don't have a horse in the game anymore for lower bounds since Tailscale is usually running close to HEAD.

.github/workflows/test.yml
Aoang: It's great without it.

Aoang:
TBR

If the minimum supported version needs to be changed, I'd be happy to help

.github/workflows/test.yml
dsnet: Oh, I see, the manual setup of dependencies were necessary folder older versions of Go before module support.

dsnet (approved):

PR #292, opened 2022-04-23
Remove xerrors dependency

catatsuy:
Versions older than Go 1.13 are no longer in use. Remove unnecessary dependencies.


dsnet:
For my @tailscale work, we only support the latest Go version. So this SGTM.

I'll leave it to @neild to make the call on what's the latest version we want to support here.

mvdan:
Oh wow, you even still test on Go 1.11.x :)

I was coming here to suggest that you drop ancient Go versions at this point, so I'm happy to see this PR. It would be nice to drop xerrors as an indirect dependency from a few of my projects.

neild (approved):

dsnet:
I think we can merge this. For some reason GitHub Actions is trying to run on `1.11` and `1.12` even though they no longer exist in the `.yml` file.

catatsuy:
Thank you!

PR #303, opened 2022-06-30
Special-case for top-level triple-quoted string literals

### LINKED ISSUE #301: Avoid unnecessary parenthesis
This is a diff I'm seeing:
```
  (
  	"""
  	... // 135 identical lines
  		"alpha@tailscale.com",
  		"bravo@tailscale.com",
+ 		"charlie@tailscale.com",
  		"delta@tailscale.com",
  		"echo@tailscale.com",
  	... // 386 identical lines
  	"""
  )
```

The surrounding `(` and `)` are redundant. Remove them:
```
  """
  ... // 135 identical lines
  	"alpha@tailscale.com",
  	"bravo@tailscale.com",
+ 	"charlie@tailscale.com",
  	"delta@tailscale.com",
  	"echo@tailscale.com",
  ... // 386 identical lines
  """
```

dsnet:
For top-level triple-quoted string literals, the surrounding parenthesis
is an unnecessary amount of verbosity and indentation.
These parenthesis are always needed for any situation where the triple-quoted
string literal does not appear at the top-level to ensure that it nests well
within some other composite structure.

Thus, we detect triple-quoted string literals at the top-level
and format them verbatim. The textList.formatExpandedTo method
assumes that it is usually called within the context of a composite structure,
so we perform some string mangling to correct the leading and trailing bytes.

Fixes #301

### DIFF
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -1125,6 +1125,18 @@ func reporterTests() []test {
 		wantEqual: false,
 		reason:    "use triple-quote syntax",
 	}, {
+		label:     label + "/TripleQuoteEndlines/MyString",
+		x:         MyString("aaa\nbbb\nccc\nddd\neee\nfff\nggg\r\nhhh\n\riii\njjj\nkkk\nlll\nmmm\nnnn\nooo\nppp\nqqq\nRRR\nsss\nttt\nuuu\nvvv\nwww\nxxx\nyyy\nzzz\n\r"),
+		y:         MyString("aaa\nbbb\nCCC\nddd\neee\nfff\nggg\r\nhhh\n\riii\njjj\nkkk\nlll\nmmm\nnnn\nooo\nppp\nqqq\nrrr\nsss\nttt\nuuu\nvvv\nwww\nxxx\nyyy\nzzz"),
+		wantEqual: false,
+		reason:    "use triple-quote syntax",
+	}, {
+		label:     label + "/TripleQuoteEndlines/Bytes",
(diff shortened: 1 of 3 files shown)


dsnet:
I'm not particularly happy with how this is implemented, but it works.

Ideally, the code change would occur in `report_slices.go` where we do the triple-quoted string literal stuff.
However, the code there fundamentally cannot implement this special-case since 
it does not know whether this is a top-level value or not.
The `defaultReporter.String` method is the place where we know we definitely have a top-level value on hand.
However, implementing it there means that we have triple-quoted related logic that is far from where it is generated.
Also, I'm not fond of the byte-mangling (i.e., with `bytes.TrimXXX`) that I needed to do.

dsnet:
I'm further questioning whether this is a good idea. It is common for someone to do something like:
```go
t.Errorf("mismatch: %s", cmp.Diff(x, y))
```
where the diff does not occur on a new line.

Previously, this would print as:
```
mismatch:   (
  	"""
  	... // 135 identical lines
  		"alpha@tailscale.com",
  		"bravo@tailscale.com",
+ 		"charlie@tailscale.com",
  		"delta@tailscale.com",
  		"echo@tailscale.com",
  	... // 386 identical lines
  	"""
  )
```
where the extra `(` and `)` would protect the diff from formatting in a strange way.

However, now it would print as:
```
mismatch:   """
  ... // 135 identical lines
  	"alpha@tailscale.com",
  	"bravo@tailscale.com",
+ 	"charlie@tailscale.com",
  	"delta@tailscale.com",
  	"echo@tailscale.com",
  ... // 386 identical lines
  """
```
where the triple-quotes a not vertically aligned.

dsnet:
I took a shower and decided this was a bad idea :)

PR #299, opened 2022-06-29
Adjust heuristic for line-based versus byte-based diffing

dsnet:
If the string has many characters that require escape sequences to print,
then we need to take that into consideration and avoid byte-by-byte diffing.

PR #312, opened 2022-11-15
Adjust heuristic for line-based versus byte-based diffing (#299)

rgosdin:
#299


dsnet:
I'm not sure what this is. Closing as it looks like an accidental PR. Feel free to send another cleaned up PR if this was intended to be a legitimate change.

PR #319, opened 2023-01-19
Fix typo error in comment

siashish:


dsnet:
Hi, thank you for the contribution, but this is written correctly as is. It is referencing the [errors.Is](https://pkg.go.dev/errors#Is) function.

siashish:
Hi @dsnet , Thanks for your feedback

PR #320, opened 2023-01-20
FIX: Remove redundant comment

siashish:


dsnet:
Thanks for the PR! This comment is not redundant. This is how the Go toolchain tests the output of example functions.

Issue #328, opened 2023-03-31
Readability of cmp.Diff function


amirrmonfared:
I believe we can make the readability of the Diff function output very much by just adding `(-want +got)`

mrwormhole:
for the love of god, please @dsnet @neild

dsnet:
I don't think this is going to change. This was discussed in the early days of `cmp` deployment, and people could not agree whether `got` or `want` comes first. People were relatively evenly split between both. In the lack of a uniform consensus, we went with an approach that avoids choosing a path.

Feel free to create a package that simply wraps `cmp.Diff` and injects the headers as you see fit.

dsnet:
For the record, I was in support of `(-got +want)`, but there were a number of notable people on the Go team who instead argued for `(-want +got)`. There are valid arguments for both:
* For `(-got +want)`, the argument is that it matches the conventional ordering of "got" appearing before "want".
* For `(-want +got)`, the argument is that it displays a diff of how the expected value was changed into the current value (i.e., what changed relative to the expected value).

The Go style guide [remains ambiguous about what direction to use](https://google.github.io/styleguide/go/decisions#print-diffs). Although, it seems to suggest `want, got`, but also seems to permit `got, want`.

neild:
And also, while "got" and "want" make sense in tests, `Diff` output isn't necessarily used only in tests.

Perhaps there should be a test-specific diff API with named parameters:

```
diff := cmp.C{
  Want: want,
  Got:  got,
}.Diff(opts)
```

mrwormhole:
very tricky, looks like it is not settled;

- https://google.github.io/styleguide/go/best-practices#local-variables-in-tests
example of cmp.Equal(got,want)
- https://pkg.go.dev/github.com/google/go-cmp/cmp#Equal
the docs of cmp.Equal(want, got) [we use this in our codebase but people are so confused which is which]

dsnet:
When it comes to `cmp.Equal` it doesn't matter since the `cmp` guarantees symmetric property of equality.

mrwormhole:
I found cmp.Diff example in google styling guide, I think it can be enforced pretty well if not suggested on README.md perhaps, I will also check out other testing libraries and come back with my findings for - + or + -

https://google.github.io/styleguide/go/best-practices#tests

but you are absolutely right equality shouldn't care about want and got, it takes 2 numbers and return bool

PR #330, opened 2023-04-21
any over interface{}

mrwormhole:
the aim of this PR is to remove support for versions below 1.18 and complete existing TODO which encourages the use of any over interface{}

~~also Diff and Equal methods in compare.go are the most commonly used functions but due to confusion of "x" and "y", it causes new developers to write want and got in reverse, so I have renamed both Diff and Equal function signatures from~~

see if this makes sense @neild @dsnet

### DIFF
--- a/cmp/internal/value/name.go
+++ b/cmp/internal/value/name.go
@@ -9,7 +9,7 @@ import (
 	"strconv"
 )
 
-var anyType = reflect.TypeOf((*interface{})(nil)).Elem()
+var anyType = reflect.TypeOf((*any)(nil)).Elem()
 
 // TypeString is nearly identical to reflect.Type.String,
 // but has an additional option to specify that full type names be used.
--- a/cmp/cmpopts/equate.go
+++ b/cmp/cmpopts/equate.go
@@ -15,7 +15,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
-func equateAlways(_, _ interface{}) bool { return true }
+func equateAlways(_, _ any) bool { return true }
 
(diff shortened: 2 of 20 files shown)


dsnet:
I'm a 👎 to renaming `got` and `want`. See my comment here: https://github.com/google/go-cmp/issues/328#issuecomment-1518120671

I'm a 👍 for only supporting higher versions of Go and using `any` over `interface{}` if @neild is okay with it. In the past, Google needed older versions internally for a long time. I no longer work at Google, and so I'm perfectly happy support only the latest 2 versions.

neild:
I'd rather wait until the x/ repos update to supporting only 1.18 before doing so here. (They're currently still on 1.17, although 1.17 is out of the supported window so they should be able to update at any time now.)

mrwormhole:
ok I will rollback this to include only `interface{}` to `any` changes. It seems like a rabbit hole discussion about want/got so I will just follow the docs example. I left a comment to the issue why I think there are inconsistencies about which to put first (want,got) or (got, want). Then I will possibly wait for x repos

cmp/internal/value/name.go
```diff
@@ -22,7 +22,7 @@ func appendTypeName(b []byte, t reflect.Type, qualified, elideFunc bool) []byte
 	// of the same name and within the same package,
 	// but declared within the namespace of different functions.
 
-	// Use the "any" alias instead of "interface{}" for better readability.
+	// Use the "any" alias instead of "any" for better readability.
```
mrwormhole, line 25: this comment can be removed if requested

Issue #136, opened 2019-04-17
Merge cmpopts and cmp package?


zombiezen:
As a semi-frequent user of the `cmp` package, most of the time I pull up godoc.org reference is to figure out whether a particular useful option is in `cmpopts` or the `cmp` package. Genuinely curious: is there a meaningful end-user benefit to keeping these packages separate? I seldom use `cmp` without `cmpopts`, and I suspect this is common.

/cc @dsnet

dsnet:
The original intention was to design a comparison library that was highly extensible with a *small* number of options. 

Implementing `cmpopts` as a separate package kept the implementation of `cmp` honest such that `cmpopts` does not rely on internal details of `cmp`. In other words, `cmpopts` is forced to be entirely implementable in terms of the first-class options provided by `cmp`. This was done in reaction to other comparison libraries I looked at that supported many options and each option deeply touched internal details of the overall comparison logic.

That said, from a end-user perspective, I can see how the two being in the same package can be convenient. At this point in time, it's a bit awkward to merge `cmpopts` into `cmp` since there would now be two ways to get at the `cmpopts` functionality.

neild:
> At this point in time, it's a bit awkward to merge cmpopts into cmp since there would now be two ways to get at the cmpopts functionality.

Presumably everything in cmpopts would be replaced with a forward to the corresponding symbol in cmp, and marked as deprecated.

I think the separation makes a lot of sense from the perspective of the implementation, but it's pretty confusing as a user. Even as a user with an above-average understanding of cmp's internals.

meling:
I came here to ask the same question as @zombiezen. I just want to point out that the `cmpopts` package isn't prominently visible from the `cmp` documentation, so as a user I spent a bit of time to find what I was looking for, and couldn't in the `cmp` package alone. When I eventually discovered the `cmpopts` package I found what I needed. 

So I would also encourage merging the two packages to make the documentation easier to use.

PR #337, opened 2023-08-26
Use of hotlinking of Go identifiers

dsnet:
GoDoc now supports hot linking to specific Go identifiers. Use that feature to enhance the usability of cmp docs.

Since cmp lacks a direct dependency on cmpopts,
you are forced to use the fully-qualified import path to references identifiers in the cmpopts package.

It can future improvements to the go/doc to handle this case, either by rendering the doc with short form,
or by detecting the use of cmpopts from the unit tests as well (since those can break circular dependencies).

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -5,7 +5,7 @@
 // Package cmp determines equality of values.
 //
 // This package is intended to be a more powerful and safer alternative to
-// reflect.DeepEqual for comparing whether two values are semantically equal.
+// [reflect.DeepEqual] for comparing whether two values are semantically equal.
 // It is intended to only be used in tests, as performance is not a goal and
 // it may panic if it cannot compare the values. Its propensity towards
 // panicking means that its unsuitable for production environments where a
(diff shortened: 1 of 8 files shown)


cmp/compare.go
```diff
@@ -66,21 +67,22 @@ import (
 //     Functions are only equal if they are both nil, otherwise they are unequal.
 //
 // Structs are equal if recursively calling Equal on all fields report equal.
-// If a struct contains unexported fields, Equal panics unless an Ignore option
-// (e.g., cmpopts.IgnoreUnexported) ignores that field or the Exporter option
-// explicitly permits comparing the unexported field.
+// If a struct contains unexported fields, Equal panics unless an [Ignore] option
+// (e.g., [github.com/google/go-cmp/cmp/cmpopts.IgnoreUnexported]) ignores that field
```
dsnet, line 71: This is extremely gross.

In my opinion, `go/doc` should also gather up related packages from the unit tests as well. That would be easiest and most concise way to fix reverse dependencies.

dsnet: Regardless, this is better than the status quo.

Alternatively, we can consider https://github.com/google/go-cmp/issues/136.

dsnet:
See https://godoc.digital-static.net/github.com/google/go-cmp@v0.0.0/cmp to get a preview of what godoc looks like after this change.

neild (approved):

PR #340, opened 2023-08-31
Add cmpopts.EquateComparable

### LINKED ISSUE #339: Default support for the `netip` package
Now that `netip` is in the stdlib, it would be nice if go-cmp supported it by default. Today you need to add:

```
cmp.Comparer(func(x, y netip.Addr) bool { return x == y }),
cmp.Comparer(func(x, y netip.Prefix) bool { return x == y }),
```

to make `cmp.Diff` work with types that include netip types.

dsnet:
This helper function makes it easier to specify that comparable types are safe to directly compare with the == operator in Go.

The API does not use generics as it follows existing options like cmp.AllowUnexported, cmpopts.IgnoreUnexported, or cmpopts.IgnoreTypes.

While generics provides type safety, the user experience is not as nice. Our current API allows multiple types to be specified:
```go
cmpopts.EquateComparable(netip.Addr{}, netip.Prefix{})
```
While generics would not allow variadic arguments:
```go
cmpopts.EquateComparable[netip.Addr]()
cmpopts.EquateComparable[netip.Prefix]()
```

Bump mininimum supported Go to 1.18 for net/netip type. Start testing on Go 1.21.

Fixes #339

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.13.x, 1.14.x, 1.15.x, 1.16.x, 1.17.x, 1.18.x, 1.19.x, 1.20.x]
+        go-version: [1.18.x, 1.19.x, 1.20.x, 1.21.x]
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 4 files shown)


dsnet:
Upgrading to Go 1.18 is not strictly necessary, if we're not ready to do that yet, I can adjust the tests accordingly.
It's still worthwhile to upgrade as that would unblock some other cleanups.

neild (approved):

PR #341, opened 2023-09-17
Do not inhibit the dead code elimination.

dominiquelefevre:
Avoid calls to value.TypeString() that do not request fully qualified type names. The implementation of TypeString() depends on getting methods by index which disables the DCE in the compiler.

k8s.io/kube-openapi and k8s.io/apiserver happen to depend on go-cmp outside tests so it is important that go-cmp does not inhibit the DCE inadvertently.

### DIFF
--- a/cmp/path.go
+++ b/cmp/path.go
@@ -168,7 +168,9 @@ func (ps pathStep) String() string {
 	if ps.typ == nil {
 		return "<nil>"
 	}
-	s := value.TypeString(ps.typ, false)
+	// NOTE: do not use value.TypeString(qualified = false) to avoid
+	// calls to reflect.Type.Method() which disables the DCE.
+	s := ps.typ.String()
 	if s == "" || strings.ContainsAny(s, "{}\n") {
 		return "root" // Type too simple or complex to print
 	}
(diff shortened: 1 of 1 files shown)


dominiquelefevre:
Hey, anybody out here? @dsnet could you please look at the patch?

dsnet:
The `cmp` package is a testing-first package, so precise printing of types is a high priority.

I'm not sure I see how this helps since `value.TypeString` is still transitively called from `cmp.Diff`.

dominiquelefevre:
https://github.com/kubernetes/kube-openapi/blob/68afd615200d5d759a9b60b66f56014d7f30494a/pkg/validation/spec/fuzz.go#L497

neild:
These both seem like cases that are better addressed by the place misusing `cmp`.

In the case of `k8s.io/kube-openapi`, I don't see why that package is taking a dependency on `cmp` just to provide the `SwaggerDiffOptions` var. There's nothing in that var that a caller can't construct themselves. If the package wants to make the `Ref` type easily comparable by `cmp`, it could provide a `Ref.Equal` method.

In the gRPC case, the use of `cmp.Equal` to compare two slices of `proto.Message` is very heavyweight. I'd replace it with a more directed comparison. (Whenever gRPC is happy requiring Go 1.21, `slices.EqualFunc` would be the simplest replacement.)

dominiquelefevre:
These are very reasonable arguments. The use in kube-openapi looks like a bug to me and I've filed them a PR. The use in gRPC, although inefficient, looks legit to me.

What bothers me more is that we are talking about a whack-a-mole solution. You've implemented some really useful features, and people will use them without thinking twice about go-cmp being mostly a package for tests. Kube-openapi, kube-apiserver and gRPC are just examples that were really trivial to find. There certainly are more. I think it is important to minimise the damage that go-cmp causes when used less than prudently. Especially given that disabled DCE seriously pessimises the code generation.

neild:
The `cmp` [documentation](https://pkg.go.dev/github.com/google/go-cmp/cmp#pkg-overview) is quite explicit that it's designed for use in tests, and is not suitable for use in production contexts:

> It is intended to only be used in tests, as performance is not a goal and it may panic if it cannot compare the values.

That line undersells "performance is not a goal", however. For example, that gRPC use of `cmp.Equal` will compare each individual list item twice, in separate goroutines, to verify that the `Comparer` is symmetric, deterministic, and does not modify its parameters.

This is quite useful for catching inadvertent errors in tests, but is really not suitable behavior for any production context.

dsnet:
If there were a build tag to isolate tests, I'd be more open to something like this, but as it stands we're going to prioritize good reporting of types in tests over reduced binary bloat.

PR #348, opened 2023-12-26
Create dependabot.yml

### LINKED ISSUE #347: Enable a dependency update tool
Hi! I'd like to suggest for go-cmp to enable a dependency update tool in order to keep CI dependencies up to date. It can also be enable to update dependencies in other ecosystems.

Having dependencies hash pinned and updated through a dep update tool is a good way to get vulnerabilities and bug fixes as soon as possible without blindly upgrading.

I'll be submiting a PR with a configuration for dependabot, but let me know if you rather renovatebot or other tool. Let me know what do you think about it.

Besides, I strongly recommend that you enable the **Dependabot security updates** option on [Code security and analysis](https://github.com/google/go-cmp/settings/security_analysis) to receive out of schedule upgrades in case of a new security patch is released (avoiding being exposed for much time). 

Disclosure: I'm from GOSST (Google Open Source Security Team) and I'm working on improving the supply security of many open source projects.

joycebrum:
Closes #347 

Here is a dependabot configuration that enables version updates for github workflows and groups them in a single PR to avoid multiple PRs

It is schedule to run monthly to allow a delay after new version bumps to allow vulnerabilities to be discovered and fixed before affecting go-cmp. Because of that, it is important to enable the "security updates" on the config, mentioned in the issue, because it enables dependabot to send out of schedule PRs in case of a security patch being released.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"


neild:
Thanks, but no thanks. `go-cmp` (quite intentionally) has no dependencies, so this isn't going to do anything useful.

PR #367, opened 2024-10-23
Support compare functions with SortSlices and SortMaps

### LINKED ISSUE #365: cmpopts: SortSlice s and SortMaps should accept func(T,T) int signature
The stdlib is gradually coalescing around `func(T, T) int` over `func(T, T) bool` as the function to compare elements. See [slices.CompareFunc](https://pkg.go.dev/slices#CompactFunc) for example.

For better or worse, the `cmpopts.SortSlices` and `cmpopts.SortMaps` uses Go reflection to call an arbitrary function (since it predates the introduction of Go generics). One benefit of the non-generic API is that we can support both a "less"-like function and a "compare"-like function.

\cc @icio

dsnet:
The SortSlices and SortMaps options predate generics and
accept an interface{}, so it is possible with reflection to support other function signatures than "func(T, T) bool".

In particular, the Go ecosystem is increasingly moving towards "func(T, T) int" as the signature for ordering as evidenced by the newer slices.SortFunc function in stdlib.

Thus, modernize cmpopts by supporting "func(T, T) int".

Fixes #365

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
 module github.com/google/go-cmp
 
-go 1.13
+go 1.21
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.18.x, 1.19.x, 1.20.x, 1.21.x]
+        go-version: [1.21.x]
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/google/go-cmp
 
-go 1.13
```
dsnet, line 3: Wow. From 4 years ago!

neild (approved):

dsnet:
CI are hosed and I don't have permissions to fix them, so you may need to merge on my behalf.

PR #370, opened 2025-01-14
Detect proto.Message types when failing to export a field

dsnet:
While we do not provide 1st class handling of protobuf messages, we can at least detect whether the type is a proto.Message and refer the user to the correct package to use to handle protobuf messages with cmp.Equal.

### DIFF
--- a/cmp/options.go
+++ b/cmp/options.go
@@ -232,7 +232,15 @@ func (validator) apply(s *state, vx, vy reflect.Value) {
 		if t := s.curPath.Index(-2).Type(); t.Name() != "" {
 			// Named type with unexported fields.
 			name = fmt.Sprintf("%q.%v", t.PkgPath(), t.Name()) // e.g., "path/to/package".MyType
-			if _, ok := reflect.New(t).Interface().(error); ok {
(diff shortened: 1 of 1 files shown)


dsnet:
Unfortunately, I can't add a test for this as doing so would require a dependency on `google.golang.org/protobuf`.

However, here's the code in action:
```go
package main

import (
	"google.golang.org/protobuf/types/descriptorpb"
)

func main() {
	m1 := &descriptorpb.DescriptorProto{}
	cmp.Equal(m1, m1)
}
```

```
panic: cannot handle unexported field at {*descriptorpb.DescriptorProto}.state:
		"google.golang.org/protobuf/types/descriptorpb".DescriptorProto
	consider using "google.golang.org/protobuf/testing/protocmp".Transform to compare proto.Message types
```

dsnet:
Note that `MethodByName` could massively increase a binary size, but `cmp` is only used for testing, so this isn't a major detriment. Also, golang/go#62257 makes it such that string literals passed to `MethodByName` are specially handled such that only those particular methods must be linked in, while the entire universe of all methods can be mostly DCE'd.

PR #371, opened 2025-02-03
Allow diff context records to be adjusted

### LINKED ISSUE #358: How to add more context to Diff output?
I like the default output of the `.Diff()` function but would love more lines of context around the differences. After looking through different examples, I can't figure out how to do this. Digging into the code of `cmp`, it looks like the context is hardcoded to `2` (https://github.com/google/go-cmp/blob/c3ad8435e7bef96af35732bc0789e5a2278c6d5f/cmp/report_compare.go#L13). Is there a way to override this value?

IdlePhysicist:
This commit enables the number context records to be adjusted, allowing for greater readability in some debugging scenarios.

Closes #358

### DIFF
--- a/cmp/report_compare.go
+++ b/cmp/report_compare.go
@@ -10,7 +10,15 @@ import (
 )
 
 // numContextRecords is the number of surrounding equal records to print.
-const numContextRecords = 2
+var numContextRecords = 2
+
+func SetNumContextRecords(n int) {
+	numContextRecords = n
+}
+
+func GetNumContextRecords() int {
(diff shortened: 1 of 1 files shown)


dsnet:
Hi, thanks for the contribution.

It's a known issue that the reporter does not print enough context for all use cases and this is something worth addressing, but I believe exposing the `numContextRecords` is too low-level of a implementation detail to expose. We probably want a single higher-level option to control the amount of printout, but it's unclear what that option should exactly be.

PR #387, opened 2026-01-28
Fix typo in comment (embedding)

manuelarte:
I was checking this library because I want to replace `reflect.DeepEqual` to `cmp.Equal` or `cmp.Diff` and while doing so, I saw that there is a typo in the word `embedding` in one of the doc comments.

I believe it has no impact at all in the code since it's just a change in a comment.

Cheers.


dsnet (approved):

manuelarte:
@dsnet anything else I need to do to merge this PR?

PR #390, opened 2026-03-11
Fix todo to replace interface{} with any

willbeason:
Resolve TODO which was set to be done after go 1.18. As the project is now in go 1.21, it seems this should be fine now.


dsnet (approved):
Thank you.

PR #391, opened 2026-03-11
Use generic function to box values on heap

willbeason:
Resolve post-1.18 todo to define a generic function that boxes a value on the heap.

### DIFF
--- a/cmp/compare_test.go
+++ b/cmp/compare_test.go
@@ -104,6 +104,7 @@ func mustFormatGolden(path string, in []struct{ Name, Data string }) {
 
 var now = time.Date(2009, time.November, 10, 23, 00, 00, 00, time.UTC)
 
+// TODO(≥go1.27): Use new(V) to allocate a value.
 func ptr[T any](t T) *T { return &t }
 
 type Stringer string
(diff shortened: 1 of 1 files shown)


willbeason:
Saw I could simplify a bit more after looking at the code in the review tool. I think I've caught all the places now that make sense to use the new generic `ptr` function.

cmp/compare_test.go
```diff
@@ -104,13 +104,11 @@ func mustFormatGolden(path string, in []struct{ Name, Data string }) {
 
 var now = time.Date(2009, time.November, 10, 23, 00, 00, 00, time.UTC)
 
-// TODO(≥go1.18): Define a generic function that boxes a value on the heap.
-func newInt(n int) *int { return &n }
+func ptr[T any](t T) *T { return &t }
```
dsnet, line 108: I would add:
```go
// TODO(≥go1.27): Use new(V) to allocate a value.
```

See https://github.com/golang/go/issues/45624

dsnet (approved):
Thank you, let's add another TODO to ironically wait for Go 1.27 support where we can use the expanded `new` semantics.

cmp/compare_test.go
willbeason: That's awesome! Added.

willbeason:
Anything else I should do for this PR?

PR #397, opened 2026-07-21
docs: clarify Equal method is skipped when either value is nil

### LINKED ISSUE #363: The documentation "Equal is called even if x or y is nil" seems incorrect.
The documentation of `Equal()` (introduced in #63) seems incorrect (emphasis mine):

> If the values have an Equal method of the form "(T) Equal(T) bool" or "(T) Equal(I) bool" where T is assignable to I, then use the result of x.Equal(y) ***even if x or y is nil***. Otherwise, no such method exists and evaluation proceeds to the next rule. 

As far as I tested, `x.Equal(y)` is NOT called if `x == nil` or/and `y == nil`.

[playground](https://go.dev/play/p/fkD4-N98-Ai)

```go
package main

import (
	"fmt"

)

type S struct {
	V *T
}

type T struct {
}

func (t1 T) Equal(t2 T) bool {
	fmt.Println("hello")
	return true
}

func main() {
	{
		s1 := S{
			V: &T{},
		}
		s2 := S{
			V: &T{},
		}
		fmt.Println(cmp.Equal(s1, s2)) //=> hello, hello, true (as expected)
	}

	{
		s1 := S{
			V: nil,
		}
		s2 := S{
			V: &T{},
		}
		// s1.V.Equal(*s2.V) //panics
		fmt.Println(cmp.Equal(s1, s2)) //=> false (doesn't panic)
	}
}
```

Solaris-star:
## Summary
Package docs for `Equal` said the method is used **even if x or y is nil**.

Actual behavior: `comparePtr` / `compareInterface` report nil equality first and return, so `tryMethod` never runs when either side is nil. Non-nil values still call `Equal` as documented.

This matches issue #363 and avoids implying a nil receiver would be invoked (and potentially panic).

## Test plan
- [x] Manual repro: nil vs non-nil `*T` with value-receiver `Equal` does not print / call method
- [x] `go test ./cmp/`

Fixes #363

### DIFF
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -58,10 +58,9 @@ import (
 //
 //   - If the values have an Equal method of the form "(T) Equal(T) bool" or
 //     "(T) Equal(I) bool" where T is assignable to I, then use the result of
-//     x.Equal(y) when both values are non-nil. If either value is nil, Equal
-//     reports whether both are nil without calling the method (so a nil
-//     receiver never panics). Otherwise, no such method exists and evaluation
-//     proceeds to the next rule.
+//     x.Equal(y). The method is called even if x or y are nil and Equal
+//     is declared on a pointer receiver. Otherwise, no such method exists and
+//     evaluation proceeds to the next rule.
 //
 //   - Lastly, try to compare x and y based on their basic kinds.
 //     Simple kinds like booleans, integers, floats, complex numbers, strings,
--- a/cmp/compare.go
+++ b/cmp/compare.go
@@ -58,8 +58,10 @@ import (
 //
 //   - If the values have an Equal method of the form "(T) Equal(T) bool" or
 //     "(T) Equal(I) bool" where T is assignable to I, then use the result of
-//     x.Equal(y) even if x or y is nil. Otherwise, no such method exists and
-//     evaluation proceeds to the next rule.
+//     x.Equal(y) when both values are non-nil. If either value is nil, Equal
+//     reports whether both are nil without calling the method (so a nil
+//     receiver never panics). Otherwise, no such method exists and evaluation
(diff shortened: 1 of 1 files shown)


cmp/compare.go
```diff
@@ -58,8 +58,10 @@ import (
 //
 //   - If the values have an Equal method of the form "(T) Equal(T) bool" or
 //     "(T) Equal(I) bool" where T is assignable to I, then use the result of
-//     x.Equal(y) even if x or y is nil. Otherwise, no such method exists and
-//     evaluation proceeds to the next rule.
+//     x.Equal(y) when both values are non-nil. If either value is nil, Equal
+//     reports whether both are nil without calling the method (so a nil
+//     receiver never panics). Otherwise, no such method exists and evaluation
+//     proceeds to the next rule.
```
dsnet: Thanks for the PR, but I don't think this is quite correct.

The difference of whether Equal is called on nil is whether it is declared on a value receiver or not. Perhaps this is more clear:

> If the values have an Equal method of the form "(T) Equal(T) bool" or
> "(T) Equal(I) bool" where T is assignable to I, then use the result of
> x.Equal(y). The method is called even if x or y are nil and Equal
> is declared on a pointer receiver. Otherwise, no such method exists and
> evaluation proceeds to the next rule.

See https://go.dev/play/p/Jf2jQPmgnZx

dsnet (changes requested):

Solaris-star:
Thanks for the correction and the playground example. I updated the package docs in commit `c1d217f` to use the suggested distinction: an `Equal` method on a pointer receiver is called even when either value is nil; the value-receiver case does not imply a callable nil receiver.

`go test ./cmp/...` passes. Could you please re-review?

PR #400, opened 2026-08-17
cmp: avoid time.Location initialization race in reports

### LINKED ISSUE #381: Race condition when using go-cmp on structs that include `time.Time`
👋 

This a follow-up to https://github.com/golang/go/issues/74460 in the golang issues tracker, where go present a data race when some of the private fields of time.Time are accessed trough the `reflect` package.

The Golang team decided not to fix the issue as it have implication on the maintainability of private fields in the standard library and suggested to fix `go-cmp` to treat `time.Time` as a special case and use `==` to compare time fields as that's the way go recommend comparing `time.Time` (read comments on https://github.com/golang/go/issues/74460 for more details).

1678092075:
## Problem

When `cmp.Diff` produces a high-verbosity report that avoids `String` methods, it can reflect over the unexported fields of a `time.Time`. If the referenced `time.Location` is being lazily initialized at the same time, the report races with the standard library's initialization writes.

This is the reporting race described in #381 and golang/go#74460.

## Root cause

The fallback report bypasses `fmt.Stringer` so it can expose otherwise hidden differences. For `time.Time`, that means recursively reading its private `Location` state without first synchronizing with lazy location initialization.

## Change

Before reflectively formatting a `time.Time` with Stringer formatting disabled, initialize the specific `Location` referenced by that value. This keeps the existing reflective output, including map-key disambiguation, while avoiding concurrent reads during lazy initialization.

The regression coverage:

- runs the race scenario in a fresh child process so the location starts uninitialized;
- replaces the global `time.Local` after constructing the value, verifying that the value's actual location is synchronized;
- verifies that distinct `time.Time` map keys with identical `String` output remain distinguishable.

## Verification

- `GOCACHE=/tmp/go-cmp-cache-regression go test -race ./cmp -run '^TestDiffDoesNotRaceWithTimeLocalInitialization$' -count=5` — pass
- `GOCACHE=/tmp/go-cmp-cache-target go test ./cmp -run '^TestFormatMapKeyDisambiguatesTimeLocations$' -count=3` — pass
- `GOCACHE=/tmp/go-cmp-cache-race go test -race ./...` — pass
- `test -z "$(gofmt -d .)"` — pass
- `git diff --check HEAD^ HEAD` — pass

## Compatibility and risk

There is no public API change. The special case only runs when report formatting intentionally bypasses Stringer methods; ordinary `time.Time` formatting and comparison behavior are unchanged.

Fixes #381.

### DIFF
--- a/cmp/report_reflect.go
+++ b/cmp/report_reflect.go
@@ -10,6 +10,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 	"unicode"
 	"unicode/utf8"
 
@@ -21,6 +22,7 @@ var (
 	stringType = reflect.TypeOf((*string)(nil)).Elem()
 	bytesType  = reflect.TypeOf((*[]byte)(nil)).Elem()
 	byteType   = reflect.TypeOf((*byte)(nil)).Elem()
+	timeType   = reflect.TypeOf(time.Time{})
 )
 
 type formatValueOptions struct {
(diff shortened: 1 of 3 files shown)


1678092075:
Hi! Just checking whether there’s anything else I can provide to help review this change. The PR is still passing the repository checks and Google CLA, and I’m happy to adjust the implementation or add coverage if needed. Thanks!

PR #404, opened 2026-09-09
.github/workflows: raise test matrix version to 1.26

randall77:
It's currently at 1.21, which no longer runs on the latest
macOS, which is what runs the tests. See
https://github.com/golang/go/issues/78012

All pull requests will fail the tests with errors like
"missing LC_UUID load command" until we upgrade to a Go version
which has the fix for 78012.


dsnet:
Thanks! Go 1.21 is pretty ancient.

dsnet (approved):

